PR #1, opened 2017-05-23
Build and library setup

MateuszKubuszok:


krzemin:
Please rename also test package.

Issue #18, opened 2017-05-26
Complete test suite for protocol buffers


krzemin:
The idea is to:
- configure [scalaPB](https://github.com/scalapb/ScalaPB) to be run only on test configuration
- define simple example protocol with (possibly) all supported PB data types - case classes should be automatically generated by scalaPB
- define corresponding domain in scala world, but with value classes instead of primitives
- write tests that check conversion round trips for many various instances

It probably includes #7 to complete (pb enums are like sealed hierarchies, but with non-total mapping, we'll need to think about it).

@MateuszKubuszok WDYT?

MateuszKubuszok:
@krzemin nifty, idea. I like it. I am afraid though that https://github.com/scalalandio/chimney/issues/19 is a blocker.

krzemin:
It was basically done.

Issue #7, opened 2017-05-23
Support for coproducts


krzemin:

krzemin:
Another important part of coproduct support was introduced with this commit:
https://github.com/scalalandio/chimney/commit/7b34f7aeebe7a22903b1d5d24f5fd53cee8358ca

@MateuszKubuszok  Do we need to support anything else?

MateuszKubuszok:
I don't think we need anything else at-the-moment. We might reconsider later on.

PR #23, opened 2017-06-01
Protocol buffer tests

krzemin:
It addresses #18 by defining simple domain model both in type-safe Scala code and Protobuf definition. For now it can derive conversion from scala to pb. There is `.proto` file provided in separate module, which scalaPB generates Scala case classes from and generated code is only used for tests.

Changes:
- added support for simple coproducts (#7), but only to handle enums by finding corresponding case objects instances. Due to SI-7046 we can't put Scala enum instances to companion object of sealed trait (but on the target side, in scalaPB generated code it can be enclosed in a such way)
- extended support for traversable collections (i.e. now we can require derived transformer from `List[A]` to `Seq[B]` assuming only that transformer from `A` to `B` exists)
- removed scala 2.12.0 from travis - due to bugs connected with SI-7046; on 2.12.1 and 2.12.2 it works fine

It still works for both JVM and JS targets.

One ugly hack was used to make things work. In GenericInstances of DerivedTransformer there is normal gen instance (with `Lazy`fied instance for generic case), but it doesn't handle some cases. OTOH, these cases are handled fine without `Lazy`, so when I put non-Lazy instance (gen2) to LowPriorityGenericInstances, all cases are magically handled! I would like to hear how to solve it properly :)

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/ValueProvider.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/ValueProvider.scala
@@ -1,6 +1,6 @@
 package io.scalaland.chimney
 
-import shapeless.{::, HList, HNil, LabelledGeneric, Lazy, Witness, ops}
+import shapeless.{::, HList, HNil, LabelledGeneric, Witness, ops}
 
 trait ValueProvider[FromG, From, T, Label <: Symbol, Modifiers <: HList] {
   def provide(from: FromG, modifiers: Modifiers): T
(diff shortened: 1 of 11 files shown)


chimney/src/main/scala/io/scalaland/chimney/ValueProvider.scala
```diff
@@ -18,12 +18,12 @@ object ValueProvider extends ValueProviderDerivation {
 
 trait ValueProviderDerivation {
 
-  implicit def hnilCase[FromG <: HList, From, FromT, ToT, L <: Symbol, Modifiers <: HNil]
+  implicit def hnilCase[Modifiers <: HNil, FromG <: HList, From, FromT, ToT, L <: Symbol]
```
MateuszKubuszok: Why moved it here?

krzemin: No strong reason, it should be equivalent, but I believe that it should speed up type inference a bit.

PR #27, opened 2017-06-05
Initial support for coproduct instances modifiers

krzemin:
Multiple modifiers that covers all the cases that uses instance `coproductInstanceCase` seems to be working.

What doesn't work yet is a coproduct type that for different cases uses both `matchingObjCase` and modifier case. It causes ambigious implicit between `matchingObjCase` and `cconsTailCase` which is weird, since the latter is in low priority trait.

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/CoproductInstanceProvider.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/CoproductInstanceProvider.scala
@@ -1,14 +1,14 @@
 package io.scalaland.chimney
 
-import shapeless.{Coproduct, HList, Witness, ops}
-import shapeless.labelled.{field, FieldType}
+import shapeless.labelled.{FieldType, field}
+import shapeless.{::, Coproduct, HList, LabelledGeneric, Witness, ops}
 
 trait CoproductInstanceProvider[Label <: Symbol, FromT, ToLG <: Coproduct, Modifiers <: HList] {
 
   def provide(src: FieldType[Label, FromT], modifiers: Modifiers): ToLG
 }
 
-object CoproductInstanceProvider {
+object CoproductInstanceProvider extends LowPriCIP {
 
   implicit final def matchingObjCase[ToLG <: Coproduct, Label <: Symbol, FromT, TargetT, Modifiers <: HList](
     implicit sel: ops.union.Selector.Aux[ToLG, Label, TargetT],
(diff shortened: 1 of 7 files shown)


krzemin:
The issue seems to be fixed now. @MateuszKubuszok feel free to review.

chimney/src/main/scala/io/scalaland/chimney/dsl.scala
```diff
@@ -32,6 +32,11 @@ object dsl {
     ): TransformerInto[From, To, Modifier.relabel[labelFrom.T, labelTo.T] :: Modifiers] =
       new TransformerInto(source, new Modifier.relabel[labelFrom.T, labelTo.T] :: modifiers)
 
+    def withCoproductInstance[Inst](
+      f: Inst => To
+    ): TransformerInto[From, To, Modifier.coproductInstance[Inst, To] :: Modifiers] =
+      new TransformerInto(source, new Modifier.coproductInstance[Inst, To](f) :: modifiers)
+
     def transform(implicit transformer: DerivedTransformer[From, To, Modifiers]): To =
```
krzemin: why?

chimney/src/main/scala/io/scalaland/chimney/CoproductInstanceProvider.scala
```diff
@@ -1,19 +1,45 @@
...
 trait CoproductInstanceProvider[Label <: Symbol, FromT, ToLG <: Coproduct, Modifiers <: HList] {
 
   def provide(src: FieldType[Label, FromT], modifiers: Modifiers): ToLG
 }
 
-object CoproductInstanceProvider {
+object CoproductInstanceProvider extends CoproductInstanceProviderDerivation {
 
-  implicit final def matchingObjCase[ToLG <: Coproduct, Label <: Symbol, FromT, TargetT, Modifiers <: HList](
+  final def provide[Label <: Symbol, FromT, To, ToLG <: Coproduct, Modifiers <: HList](
```
krzemin: I agree. We can improve naming as well, but maybe in a separate commit/PR?

MateuszKubuszok: Sure, just a cleanup thing, we can do it after merge.

chimney/src/main/scala/io/scalaland/chimney/dsl.scala
MateuszKubuszok: Inside internals order should not matter as much... but it would "feel" better for future users that might want to implement `implicit val t: Transformer[...] = ...` when whatever they are impementing matches the signature of a `.transform` call - otherwise we have a situation when user creates `Transformer` and method accepts `DerivedTransformer`. I think that potentially that might cause a trouble to users. But that's for the future to discuss.

krzemin: I think using DerivedTransformer is crucial here because unlike Transformer, it respects modifiers. For the user it's more important to have a good documentation that explains how to use the library rather than deeply understand design decisions.

krzemin: I think using DerivedTransformer is crucial here because unlike Transformer, it respects modifiers. For the user it's more important to have a good documentation that explains how to use the library rather than deeply understand design decisions.

Issue #31, opened 2017-06-22
Cross-publish for Scala 2.11.9 and later


PaulAtBanno:
Since [SI-7046 is resolved in Scala 2.11.9 or later](https://issues.scala-lang.org/browse/SI-7046), could you please cross-publish for it?

krzemin:
Yes, that should be easily possible. Probably the only thing we have to do before is to replace instances written in SAM-style to portable ones.

krzemin:
Version 0.1.4 is published for 2.11 as well.

PR #32, opened 2017-06-22
Add support for Scala 2.11

krzemin:
This PR adds support for Scala 2.11, which was requested in #31.
It contains implicit instances rewritten from SAM-style to style with helper instance creators, which is supported also in 2.11. Due to explicit instances for `Left/Right` I had to temporarily disable deprecation checking as it was causing build failure. Later we may consider using silencer plugin to have more fine grained control over scoped deprecation warnings.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,7 @@ val settings = Seq(
     "-encoding",
     "UTF-8",
     "-unchecked",
-    "-deprecation:false", // just for the time being
+    "-deprecation",
     "-explaintypes",
     "-feature",
     "-language:existentials",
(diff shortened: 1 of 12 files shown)


build.sbt
```diff
@@ -1,12 +1,13 @@
 val settings = Seq(
   version := "0.1.3",
   scalaVersion := "2.12.2",
+  crossScalaVersions := Seq("2.11.11", "2.12.2"),
   scalacOptions ++= Seq(
     "-target:jvm-1.8",
     "-encoding",
     "UTF-8",
     "-unchecked",
-    "-deprecation",
```
MateuszKubuszok: Could you add the reason? Just to keep the trace "why?".

krzemin: I have an idea how to fix it by using extractor instead of field selection.

Issue #37, opened 2017-08-21
Optional patcher


piotrkosecki:
Hi. Can you provide something like optional patcher? The idea would be to do something like this:
```
case class Foo(a: String, b: Int, c: Double)
case class Bar(a: Option[String], b: Option[Int])

val f = Foo("aaa", 123, 0.33)
val b = Bar(None, Some(987))
f.patchWithOpt(b) should be Foo("aaa", 987, 0.33) 
```
At the moment the way I can do it is: 
```
f.copy(a = b.getOrElse(f.a), ...) 
```

PR #38, opened 2017-08-27
Support for optional fields in patchers

krzemin:
This PR addresses #37 proposal.

It doesn't provide separate method in DSL, but integrates (I believe) clearly with current approach. 

So far, when we had a patch field field `f` of type `T`, we required a field `f` of type `U` to be present in target object, and transformation from `T` to `U`.

This PR adds a rule, that beside this one above, supports also fields `f` of type `Option[T]` in patch, and requires transformation from `T` to `U`.

See the test case for reference.

### DIFF
--- a/chimney/src/test/scala/io/scalaland/chimney/PatcherSpec.scala
+++ b/chimney/src/test/scala/io/scalaland/chimney/PatcherSpec.scala
@@ -33,9 +33,9 @@ class PatcherSpec extends WordSpec with MustMatchers {
 
       import TestDomain._
 
-      case class UserPatch(email: Option[Email], phone: Option[Phone])
+      case class UserPatch(email: Option[String], phone: Option[Phone])
 
-      val update = UserPatch(email = Some(Email("updated@example.com")), phone = None)
+      val update = UserPatch(email = Some("updated@example.com"), phone = None)
 
       exampleUser.patchWith(update) mustBe
         User(10, Email("updated@example.com"), Phone(1234567890L))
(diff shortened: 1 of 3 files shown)


chimney/src/test/scala/io/scalaland/chimney/PatcherSpec.scala
```diff
@@ -23,12 +23,23 @@ class PatcherSpec extends WordSpec with MustMatchers {
...
 
-      val user = User(10, Email("abc@def.com"), Phone(1234567890L))
       val update = UpdateDetails("xyz@def.com", 123123123L)
 
-      user.patchWith(update) mustBe
+      exampleUser.patchWith(update) mustBe
         User(10, Email("xyz@def.com"), Phone(123123123L))
     }
+
+    "support optional types in patch" in {
```
piotrkosecki: does it support both optional and regular types? i.e. if I had such test:
```
"support mixed optional and regular types" in {
    import TestDomain._

    case class UserPatch(email: String, phone: Option[Phone])
    val update = UserPatch(email = "updated@example.com", phone = None)
    
    exampleUser.patchWith(update) mustBe User(10, Email("updated@example.com"), Phone(1234567890L))
}
```

krzemin: Yes, I expect this test should pass. I will include it to the PR later.

MateuszKubuszok:
:+1:

chimney/src/test/scala/io/scalaland/chimney/PatcherSpec.scala
krzemin: Test added. It works without problems, as expected.

Issue #40, opened 2017-09-11
Simple example failure


zodo:
Hello
```
scala> case class One(text: Option[String])
scala> case class Two(text: Option[String])
scala> val one = One(None)
scala> import io.scalaland.chimney.dsl._
scala> one.transformInto[Two]
<console> error: could not find implicit value for parameter derivedTransformer: io.scalaland.chimney.DerivedTransformer[One,Two,shapeless.HNil]
             one.transformInto[Two]
                              ^
```
Scala version: 2.12.3,
Library version: 0.1.5
Any advice?

krzemin:
@zodo thanks for the bug report. There were ambiguous transformer instances for this specific case. Fixed in #41.

krzemin:
There were similar issues concerning other types that we provide particular instances for (like either, collections, maps). All of them were addressed in #41.

Closing, as it's now merged.

PR #41, opened 2017-09-11
Fix for Option[T] to Option[T] transformation

krzemin:
This PR contains proposal for fixing #40. There was ambiguity between identity transformer instance and `Option[From]` to `Option[To]` one. This solution adds requirement on the latter instance that `From` has to be different than `To`.

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/DerivedTransformerInstances.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/DerivedTransformerInstances.scala
@@ -33,7 +33,8 @@ trait ValueClassInstances {
 trait OptionInstances {
 
   implicit final def optionTransformer[From, To, Modifiers <: HList](
-    implicit innerTransformer: DerivedTransformer[From, To, Modifiers]
+    implicit neq: From =:!= To,
+    innerTransformer: DerivedTransformer[From, To, Modifiers]
   ): DerivedTransformer[Option[From], Option[To], Modifiers] =
     (src: Option[From], modifiers: Modifiers) => src.map(innerTransformer.transform(_, modifiers))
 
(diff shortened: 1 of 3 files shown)


chimney/src/main/scala/io/scalaland/chimney/internal/DerivedTransformerInstances.scala
```diff
@@ -51,7 +52,8 @@ trait OptionInstances {
 trait EitherInstances {
 
   implicit final def eitherTransformer[FromL, ToL, FromR, ToR, Modifiers <: HList](
-    implicit leftTransformer: DerivedTransformer[FromL, ToL, Modifiers],
+    implicit neq: (FromL, FromR) =:!= (ToL, ToR),
```
MateuszKubuszok: How will it behave when one of types is changed and other is kept?

chimney/src/main/scala/io/scalaland/chimney/internal/DerivedTransformerInstances.scala
```diff
@@ -80,21 +82,24 @@ trait EitherInstances {
...
-    implicit innerTransformer: DerivedTransformer[From, To, Modifiers],
+    implicit neq: From =:!= To,
+    innerTransformer: DerivedTransformer[From, To, Modifiers],
     toTag: ClassTag[To]
   ): DerivedTransformer[Array[From], Array[To], Modifiers] =
     (src: Array[From], modifiers: Modifiers) => src.map(innerTransformer.transform(_: From, modifiers))
 
   implicit final def mapTransformer[FromK, ToK, FromV, ToV, Modifiers <: HList](
-    implicit keyTransformer: DerivedTransformer[FromK, ToK, Modifiers],
+    implicit neq: (FromK, FromV) =:!= (ToK, ToV),
```
MateuszKubuszok: Same question as with `Either`?

Issue #21, opened 2017-05-26
Usage of tags can break derivation


MateuszKubuszok:
Test case:

```
object TestConversion {
  import io.scalaland.chimney.dsl._
  import shapeless.tag
  import shapeless.tag._

  sealed trait Tag

  final case class Input(
    long:   Long @@ Tag,
    string: String
  )

  final case class Output(
    long: Long @@ Tag
  )

  val test = Input(tag[Tag](0L), "").transformInto[Output]
}
```

results in `Could not find implicit for derivedTransformer...`

MateuszKubuszok:
I we remove tag OR if we remove `string` param OR if we replace `long` with `string` in `Output` class derivation succeeds.

MateuszKubuszok:
@krzemin investigated the issue and found out it is a problem described in https://github.com/milessabin/shapeless/pull/726 - if we figured that one out we could fix the issue.

krzemin:
If this is the only dependent issue, then it should work under 2.11. Unfortunately, we don't support 2.11 atm :)

krzemin:
As we now support 2.11, I tried it and confirm it works on 2.11 with no problems.

PR #42, opened 2017-09-11
fix issue #21 by updating shapeless

krzemin:
As https://github.com/milessabin/shapeless/pull/761 was merged to shapeless, I've decided to try updating to the newest snapshot version. It seems to fix the issue #21 on Scala 2.12+.

But I'd prefer not to depend on snapshot version and wait until shapeless 2.3.3 is released.


krzemin:
As shapeless 2.3.3 is released, I guess we can proceed with that.

PR #43, opened 2017-09-21
Use sbt 1.0.2

krzemin:
Finally I was able to configure build that uses sbt 1.0.2 and all needed plugins (scalapb, scoverage, scalafmt). I had to downgrade scalapb to 0.6.0, but it seems to work seamlessly now.


MateuszKubuszok:
LGTM :+1:

Issue #46, opened 2017-09-28
withFieldConst doesn't for work None


krzemin:
This doesn't work, but should!

```scala
case class X(a: Int)
case class Y(a: Int, b: Option[String])

X(5).into[Y].withFieldConst(_.b, None).transform

error: could not find implicit value for parameter transformer: io.scalaland.chimney.DerivedTransformer[X,Y,io.scalaland.chimney.internal.Modifier.fieldFunction[Symbol with shapeless.tag.Tagged[String("b")],X,None.type] :: shapeless.HNil]
       X(5).into[Y].withFieldConst(_.b, None).transform
                                              ^
```

when `None` is explicitly up-casted to `Option[String]`, it works seamlessly:

```scala
X(5).into[Y].withFieldConst(_.b, None : Option[String]).transform
```

zarthross:
`Option.empty[String]` also works, and looks a bit better than casting.

krzemin:
Closing, as #49 is merged.

PR #49, opened 2018-01-07
Fix issue #46 by lubbing modifier return type

zarthross:


zarthross:
The failed Codacy Quality review is from an existing issue that appears to be a false positive.  I thinks we are using `head` on a scala collection, but we are using `head` on an hlist which is perfectly fine.

PR #58, opened 2018-04-18
Make Modifier classes public.

rpiaggio:
This PR allows users to greate generic methods that call `_.transformInto[A]`.

Using Diode, I found myself writing a lot of `SPACircuit.connect(_.transformInto[...])` and when trying to abstract away the whole call (in order to have even less boilerplate) by creating a method:

```
def connect[A <: AnyRef](implicit derivedTransformer: DerivedTransformer[RootModel, A, Modifier.enableDefaultValues :: HNil]) = SPACircuit.connect(_.transformInto[A])
```

I found out it wouldn't compile because the type `Modifier.enableDefaultValues` was private.

This change allows further levels of abstraction on top of `chimney`, allowing intermediate methods to be parametrized or not on the `From` and `To` types.

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/Modifier.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/Modifier.scala
@@ -6,10 +6,10 @@ sealed trait Modifier
 
 object Modifier {
 
-  private[chimney] class fieldFunction[Label <: Symbol, From, T](val map: From => T) extends Modifier
-  private[chimney] class relabel[FromLabel <: Symbol, ToLabel <: Symbol] extends Modifier
-  private[chimney] class coproductInstance[From, T](val f: From => T) extends Modifier
-  private[chimney] class enableDefaultValues extends Modifier
+  class fieldFunction[Label <: Symbol, From, T](val map: From => T) extends Modifier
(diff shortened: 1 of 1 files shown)


krzemin:
Hey @rpiaggio,  

thank you for the PR and raising the issue.

Modifiers were intentionally hidden as private classes as they are internal things of the library and may change in the future (and most likely will). I'm not very eager to open them to public, instead I'd like to analyze your use case and perhaps find better way to solve it, than using undocumented internals.

If you don't use default values in your models, you can try with just passing `HNil` and it should work.

The intended solution that I'm working on is to have derivation working on `Transformer` type class, so that you could write:

```
def connect[A <: AnyRef](implicit t: Transformer[RootModel, A]) = SPACircuit.connect(t.transform)
```

Hope it will be available in 0.2.x. Stay tuned :)

rpiaggio:
Thank you @krzemin! That looks great and definitely looks cleaner in the code. I'll eagerly await 0.2 then.

krzemin:
@rpiaggio FYI: version `0.2.0-SNAPSHOT` was today released to Sonatype repo. You can try out whether it meet your use case.

rpiaggio:
Thank you very much @krzemin! Works like a charm and the API is more clear.

PR #70, opened 2018-07-22
Fixes #67: Support any class as source as long as target is a case class

### LINKED ISSUE #67: Support reading from non-case classes
This would be really convient in a ScalaJs context.  Native JS classes defined in Scala can't be `case` classes.  Frequently, I want to transform native JS classes to my own internal case classes in order to use code shared with the JVM.

Obviously I would prefer reading and writing but I assume reading is much easier since I think you can enumerate all of the class fields.

asdcdow:


chimney/src/test/scala/io/scalaland/chimney/DslSpec.scala
```diff
@@ -531,6 +531,39 @@ class DslSpec extends WordSpec with MustMatchers {
...
+
+        target.id mustBe source.id
+        target.name mustBe source.name
+      }
+
+      "support java beans" in {
+        val source = new JavaBeanSource("test-id", "test-name")
+        val target = source
+          .into[CasesTarget]
+          .withFieldRenamed(_.getId, _.id)
```
krzemin: If we really want to support reading from Java bean sources, I think we should interpret `.getXXX` methods automatically. But I think it's separate issue: https://github.com/scalalandio/chimney/issues/71

PR #73, opened 2018-08-14
Add _root_.chimney package object with aliases

### LINKED ISSUE #72: Change base package to chimney?
With syntax-heavy libraries it usually helps a lot to have a shorter implicit. Currently with Chimney I find myself doing `import io.scalaland.chimney.dsl._` a lot, but what I'd really like to do is just `import chimney.implicits._` or `chimney.dsl._`.

I suggest renaming the base package in a future release (maybe by introducing a type alias to the old package, or move everything and have deprecated type aliases pointing to the new package).

kubukoz:
Fixes #72.

For now I decided to go with the option of creating a new package object and aliasing the members of `io.scalaland.chimney` there - perhaps it's a better idea to leave them there than using a top-level chimney package. Feel free to suggest using one of the alternatives :)

### DIFF
--- a/chimney/src/main/scala/chimney/package.scala
+++ b/chimney/src/main/scala/chimney/package.scala
@@ -1,0 +1,11 @@
+/**
+  * Aliases for easier access to the Chimney API.
+  * */
+package object chimney {
+  import io.scalaland.{chimney => ch}
+
+  val dsl: ch.dsl.type = ch.dsl
+
+  type Patcher[T, Patch] = ch.Patcher[T, Patch]
+  type Transformer[From, To] = ch.Transformer[From, To]
+}
--- a/build.sbt
+++ b/build.sbt
@@ -140,7 +140,7 @@ lazy val readme = scalatex
     url = "https://github.com/scalalandio/chimney/tree/master",
     source = "Readme"
   )
-  .settings(noPublishSettings : _*)
+  .settings(noPublishSettings: _*)
   .settings(
     scalaVersion := versions.scalaVersion,
     siteSourceDirectory := target.value / "scalatex",


chimney/src/main/scala/chimney/package.scala
```diff
@@ -0,0 +1,11 @@
...
+  * Aliases for easier access to the Chimney API.
+  * */
+package object chimney {
+  import io.scalaland.{chimney => ch}
+
+  val dsl: ch.dsl.type = ch.dsl
+
+  type Patcher[T, Patch] = ch.Patcher[T, Patch]
+  type Transformer[From, To] = ch.Transformer[From, To]
+}
```
krzemin: I believe we need to alias also companion objects of `Patcher` and `Transformer` types. For instance one could use `Patcher.derive` or `Transformer.derive` in order to run derivation macro. Without these aliases it wouldn't be possible.

krzemin:
Wouldn't it be better to move everything to short `chimney` package and leave (deprecated) type aliases in long `io.scalaland.chimney` just for compatibility?

Would you mind adding some tests that verifies both import paths actually work?

chimney/src/main/scala/chimney/package.scala
kubukoz: Right, for some reason I thought they were implicits and nobody would call them directly - not true :P

kubukoz:
Yeah, I'll move everything then. A few tests too

chimney/src/main/scala/chimney/package.scala
krzemin: I would mainly expect such usage too but since these symbols are public, we can't assume nobody uses them directly.

PR #74, opened 2018-08-14
Add coursier

kubukoz:


tues:
Isn't it better to enable Coursier globally in your `~/.sbt/` instead?

kubukoz:
Maybe. But it won't work in CI :P

tues:
It'll be a very unpleasant surprise when someone clones this project, runs `sbt` and it suddenly starts to re-download all JVM libraries commonly used in the Scala ecosystem. I'm sure there are better ways to enable Coursier in CI... I might try doing that later today.

kubukoz:
Sure, go ahead. I don't know how to do that :P

krzemin:
We were using coursier some time ago (it was one of the 1.0.0-RCxx version), but it was causing random problems on CI so that I went back to sbt.

PR #77, opened 2018-08-27
Scala 2.13 support

krzemin:
As Scala 2.13 is going to be released soon, we need to cross support 2.13 too. I tried to play with latest available milestone (`2.13.0-M4`) and it seems that number of required changes in quite short, most of them caused by new standard library.

##### Deprecation of `Traversable`
We may use `Iterable` as well, covering all of the use cases. This doesn't even need special changes to 2.11/2.12 codebase.

##### No support for `collection.to[AnotherCollection]`
But `collection.to(AnotherCollection)` works great. We need to generate different code, depending on Scala version to keep the same semantics. But in 2.13 we need to use companion ref hack...

##### Unsupported compiler options
In 2.13.0-M4 it seems following two flags are not understood by scalac: `-Yno-adapted-args`, `-Xexperimental`.

##### Dependencies
Before merging, we need to wait for following dependencies:
- ~~[ ] scalatest (already published snapshot `3.0.6-SNAP2` for `2.13.0-M4`)~~ switched to utest
- ~~[ ] scalapb~~ not dependent on scalapb any more
- ~~[ ] scalatex~~ not needed for cross compilation
- [X] scoverage (bumped `sbt-scoverage` to `1.6.0-M5`)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: scala
 scala:
    - 2.11.12
-   - 2.12.6
    - 2.12.7
+   - 2.13.0-M5
 jdk:
   - openjdk8
   - openjdk11
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,7 @@ val versions = new {
 val settings = Seq(
   version := "0.2.1",
   scalaVersion := versions.scalaVersion,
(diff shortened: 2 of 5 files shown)


krzemin:
Finally dropped scalatest, scalapb and bumped versions of few libraries. I think now it's ready to merge.

Issue #47, opened 2017-09-28
Better handling Option[T] in target when no value provider can be found


krzemin:
Current support for `Option[T]` types is IMO not sufficient enough. For now we support simple transformation derivation rule that having transformation `A -> B` can infer transformation of `Option[A] -> Option[B]`. It's useful, but doesn't handle one common case, especially present in data types evolution:

```scala
case class Foo1(x: String)
case class Foo2(x: String, y: Option[Int])
```

We can't automatically transform `Foo1` to `Foo2` (and probably we shouldn't by default), but it would be nice to have a modifier that automatically handles all optional values in a way that in case when no provider in the source type can be found, we can fill it automatically with `None`.

Example usage:

```scala
Foo1("abc").into[Foo2].withMissingOptionalsDefaultToNone.transform
```

Obviously, shorter name than `withMissingOptionalsDefaultToNone`, but still being able to express the intent would be useful.

It should be probably addressed after fixing #46.
We should check how it interacts with #45 proposal. Maybe we don't need both of them?

krzemin:
Thank you @jgogstad!

PR #80, opened 2018-09-09
New configuration: .disableOptionDefaultsToNone

jgogstad:
I stuck with the "enabling" semantics in https://github.com/scalalandio/chimney/issues/47.

The code quickly gets confusing with double negations if we go with something like ".disableOptionDefaultsToNone". Like `if (!disableOptionDefaultsToNone)`, but if you're ok with it, I'll make the change.

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/TransformerMacros.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/TransformerMacros.scala
@@ -415,7 +415,7 @@ trait TransformerMacros {
         .orElse {
           val targetTypeIsOption = targetField.returnType <:< typeOf[Option[_]]
           if (targetTypeIsOption && config.optionDefaultsToNone) {
-            Some(ResolvedFieldTree(q"scala.None"))
+            Some(ResolvedFieldTree(q"_root_.scala.None"))
           } else {
             None
           }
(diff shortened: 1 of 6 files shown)


chimney/src/main/scala/io/scalaland/chimney/internal/TransformerMacros.scala
```diff
@@ -412,6 +412,14 @@ trait TransformerMacros {
             None
           }
         }
+        .orElse {
+          val targetTypeIsOption = targetField.returnType <:< typeOf[Option[_]]
+          if (targetTypeIsOption && config.optionDefaultsToNone) {
+            Some(ResolvedFieldTree(q"scala.None"))
```
krzemin: `_root_.scala.None` would be better

krzemin:
Note that your configuration entry (on the type level or value level) doesn't necessary need to follow semantics of the dsl operator - I'm ok with having them negated so that is dsl you have `.disableOptionDefaultsToNone ` and in the configuration `optionDefaultsToNone`.

jgogstad:
Has the codacy failure been around for a while, or did I introduce that?

krzemin:
Regarding Codacy, we have test coverage of ~89%, but it fails if the PR decreases the previous value. But I see you have already added one more case to prevent that 👍

krzemin:
@Krever I'd like to discuss what defaults are sensible and how we should configure it in a separate issue: https://github.com/scalalandio/chimney/issues/85. For now, just merging as it is.

PR #82, opened 2018-09-09
Rename w/ transform? Fixed #76

### LINKED ISSUE #76: Rename w/ transform?
I was thinking it might be nice to have a version of `withFieldRenamed` that would rename + take an in-scope `Transformer` to rename and transform at the same time; currently you would need to do `withFieldComputed`

e.g.

```scala
case class InternalTicket(createdAt: DateTime)

case class ExternalTicket(createdAtMillis: Long)

implicit val jodaDateTimeToLong = new Transformer[DateTime, Long] {
  def transform(dt: DateTime): Long = dt.toInstant.getMillis
}

val externalTicket = InternalTicket(DateTime.now).into[ExternalTicket]
  .withFieldRenamed(_.createdAt, _.createdAtMillis)
```

Currently, it seems like we need to do

```scala
val externalTicket = InternalTicket(DateTime.now).into[ExternalTicket]
  .withFieldComputed(_.createdAtMillis, _.toInstant.getMillis)
```

which is marginally longer for this small example, but can get repetitive if there are more fields.

I also tried with an in-scope implicit conversion, but it doesn't get picked up.

As with my previous issue, it might be that this is already possible, but it doesn't seem like it 😅

wenta:


krzemin:
I think the only thing left is scalafmt.

Issue #90, opened 2018-11-08
Add default transformation from T to Option[T]


Krever:
```
import io.scalaland.chimney.Transformer

trait ChimneyUtils {

  implicit def someByDefault[T]: Transformer[T, Option[T]] = (x: T) => Some(x)

  implicit def getByDefault[T]: Transformer[Option[T], T] = (x: Option[T]) => x.get

}
```

I wouldnt enable `getByDefault` byDefault but the other one is rather safe.

krzemin:
Transformation `T => Option[T]` is implemented in #94 and enabled by default.

Regarding second rule proposal, this was originally requested in #62. I'm rather reluctant to provide builtin support for unsafe operations, especially when it can be achieved by defining single implicit def.

PR #94, opened 2018-11-23
Support transformation of T to Option[T]

krzemin:
It addresses #90.


lloydmeta:
Lightning fast turn around. Thanks !

Issue #64, opened 2018-06-17
Support for custom strategy of patching via options


kubukoz:
EDIT: The actual description is in @krzemin's post below. This was originally a different issue.

Given
```scala
case class Foo(name: String, age: Int)
case class FooPatch(name: Option[String], age: Option[Int])
```

there could be a function taking instances of both and modifying the `Foo` with the present fields from `FooPatch`.

krzemin:
Currently optional fields are supported, but when the optional field in patch is non-empty, it takes precedence over existing field. As I understand, this proposal is to customize behavior to be able to combine final value from the two existing.

We need to find good api for this. I'd like to keep existing behavior default and allow to override by providing the function.

Because existing `patchWith` performs patching in place, we don't really have a space to provide a function to patcher. But following the same convention as we have for transformers (having `transformInto` vs `into[T].transform` in case we need to provide something extra), we could have `with[P].patch`-like syntax.

Then the example code that utilizes the generalized patching feature could look like: 

```
foo
  .with(fooPath)
  .combine(_.name, (old: String, new: String) => old + new)
  .combine(_.age, (old: Int, new: Int) => Math.max(old, new))
  .patch
```

I think this combining mechanism could equally work for both optional and non-optional fields.

kubukoz:
Actually, I was asking about the functionality that you said already exists :P it seems to be left out of the documentation, that's why I had to ask. But providing a custom way for selecting the value to use sounds neat.

krzemin:
Yes, we're missing that in docs. I created #65 then.

PR #98, opened 2019-01-29
Allow extra fields in Patcher classes that don't exist in original

codingismy11to7:
There was a specific check put in to guard against this behavior, but no tests ensuring this was the behavior - so thought maybe it could be changed.

Example scenario: we have an api, and want to update a user. The user can specify any of the user fields to update. I'd like to use most of the fields to patch the user, but there's some that need massaging first.

```scala
case class User(name: String, email: String, roleIds: Set[RoleId])
case class UserUpdate(name: Option[String], email: Option[String], roleNames: Option[Set[String]])
```

In this simple case, I want to use the `UserUpdate` class to patch the `User` class, even though I have to handle the verification/conversion/patching of `roleNames` into `roleIds` myself.

It may be that this should be handled through some flag or some enhanced patch api with transformers or something, just wanted to start the discussion.

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/PatcherMacros.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/PatcherMacros.scala
@@ -48,15 +48,15 @@ trait PatcherMacros {
       val fnObj = c.internal.reificationSupport.freshTermName("obj$")
       val fnPatch = c.internal.reificationSupport.freshTermName("patch$")
 
-      val targetMapping = patchParams.toSeq.map { pParam =>
-        tParamsByName.get(pParam.name) match {
-          case Some(tParam) if pParam.returnType <:< tParam.returnType =>
-            Right(q"$fnPatch.${pParam.name}")
-          case Some(tParam) =>
-            transformOptionalValue(fnPatch, pParam, tParam, fnObj)
-          case None =>
-            Left(s"Field named '${pParam.name}' not found in target patching type $T!")
-        }
+      val matchingParams = patchParams.toSeq flatMap { pParam =>
+        tParamsByName.get(pParam.name).map(_ -> pParam)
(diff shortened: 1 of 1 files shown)


codingismy11to7:
To answer the last question first, in my specific case, renamed fields aren't crucial.

The proposal in #64 does seem very useful, and could be used for my particular case (although i'd be throwing away the `new` value because my `Set[String] -> Set[RoleId]` conversion is in a cats `Validated`, and by the time i've gotten to the patching step i've already computed the role id set). Regardless, anything that could work would be good.

Looking at it again, I'm assuming the types would have to be the same, so that might not work for my case after all. `RoleId` is not a value class in my case (was just making an example), it's a `UUID` that i've mapped from the (`String`) name given by the end user.

Anyway, not sure about exact details, but if types could be different or there was to be a new patching syntax and we could mark fields as ignored, I could work with it

krzemin:
> I'm assuming the types would have to be the same, so that might not work for my case after all

I assumed types don't need to be literally equal, but there exists chimney transformation for them. But you're right in your case it would not work (unless you put your custom mapping logic into implicit transformer in scope which is... meh, not the best idea :P).

So we need to find an API that is able to express really custom logic.

What about this one?

```scala
user
  .with(userPatch)
  .enableIncompletePatches // don't emit compile error when you find a field that doesn't match the target
  .withFieldValue(_.roleIds, (old: Set[RoleId], patch: UserPatch) => ...)
  .patch
```

codingismy11to7:
Sorry - got busy.

That proposal looks great! I think it would solve all sorts of use cases.

krzemin:
Incomplete patches support was merged in #130.
For custom logic part I created #134.

lokolte:
@krzemin can you provide the imports for this example?, i couldn't find them.

Issue #105, opened 2019-04-24
Ability to use dsl inside Transformer definition


REDNBLACK:
I want to define transformation logic in separate objects, but i also want to use transformation dsl inside of Transformer definition. Currently this piece of code gives StackOverflow.

```
import io.scalaland.chimney.dsl._
import io.scalaland.chimney.Transformer

object Run extends App {
  case class Foo(a: String, b: Int, c: Int)
  case class Bar(a: String, b: Int, x: Long)

  object TransformerInstances {
    implicit val fooBarTransformer: Transformer[Foo, Bar] = foo => {
      foo.into[Bar]
        .withFieldComputed(_.x, _.c.toLong)
        .transform
    }
  }

  import TransformerInstances._

  val foo = Foo("a", 1, 3)

  println(foo.transformInto[Bar])
}

```

krzemin:
I am aware of this issue, but I don't know whether it's possible to resolve it properly in the Chimney's implementation.

Briefly, `.transform` call in your instance requires implicit instance of `Transformer[Foo, Bar]` which may be automatically generated by macro, or supplied through implicit scope. In this case, Scala compiler picks `implicit val fooBarTransformer` which is being just defined.

You can work it around by juggling implicit instances, like:

```scala
object TransformerDefinitions {
   val fooBarTransformer: Transformer[Foo, Bar] = foo => {
      foo.into[Bar]
        .withFieldComputed(_.x, _.c.toLong)
        .transform // no implicit Transformer[Foo, Bar] in scope, macro goes in!
    }
}

object TransformerInstances {
  implicit val fooBarTransformer: Transformer[Foo, Bar] =
    TransformerDefinitions.fooBarTransformer // introduce already generated transformer into implicit scope here
}
```

nadavwr:
In Circe it is possible to use a derived `Decoder` in order to define an implicit `Decoder` manually.
This is how they do it:

```scala

case class Exported[+T](instance: T) extends AnyVal

trait LowPriorityDecoderInstances {
  implicit final def importedDecoder[A](
    implicit exported: Exported[Decoder[A]]
  ): Decoder[A] = exported.instance
}

// derivation actually creates `Exported[Decoder[A]]`
trait AutoDerivation {
  implicit def exportDecoder[A]: Exported[Decoder[A]] = macro ExportMacros.exportDecoder[DerivedDecoder, A]
[...]

PR #112, opened 2019-07-08
Alternative solution of #105 issue

ulanzetz:
If we will forbid .into[B].transform and use only .transformTo[B] to direct transformation, we will able to use transformation DSL inside of Transformer definition.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,7 +5,7 @@ val versions = new {
 }
 
 val settings = Seq(
-  version := "0.3.2",
+  version := "1.0.0",
   scalaVersion := versions.scalaVersion,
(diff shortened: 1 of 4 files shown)


krzemin:
Thank you for your effort with providing solution to this problem. The idea is pretty fine, but I don't like changing default behavior of `into[T].transform` and introducing semantic inconsistency between this and `.transformInto[T]`.

It seems this is not inherently required. I prepared proposal based on your idea that maintains backwards compatibility. Please take a look at #114 and tell what do you think.

MateuszKubuszok:
Closed in favor of https://github.com/scalalandio/chimney/pull/129 - thanks for contribution!

PR #114, opened 2019-07-09
Enable to use dsl inside transformer being defined

krzemin:
This PR is one another proposal that addresses #105. The idea is pretty much the same as in #112, but this one:
- maintains backwards compatibility of `.into[T].transform`
- maintains semantical equivalence between `.into[T].transform` with `.transformInto[T]`
- doesn't introduce any wrapper types for exported definitions
- allows to manually disable top-level implicit lookup by using new modifier `.disableLocalImplicitLookup`
- provides alternative, shorter syntax `defineInto[T]`

@ulanzetz, @REDNBLACK - I'd be thankful for review and feedback.

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/dsl.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/dsl.scala
@@ -11,6 +11,9 @@ object dsl {
     final def into[To]: TransformerInto[From, To, Empty] =
       new TransformerInto[From, To, Empty](source, Map.empty, Map.empty)
 
+    final def defineInto[To]: TransformerInto[From, To, EnableOptionDefaultsToNone[Empty]] =
+      new TransformerInto[From, To, EnableOptionDefaultsToNone[Empty]](source, Map.empty, Map.empty)
+
     final def transformInto[To](implicit transformer: Transformer[From, To]): To =
       transformer.transform(source)
   }
--- a/chimney/src/main/scala/io/scalaland/chimney/dsl.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/dsl.scala
@@ -31,6 +31,9 @@ object dsl {
     def enableOptionDefaultsToNone: TransformerInto[From, To, EnableOptionDefaultsToNone[C]] =
       new TransformerInto[From, To, EnableOptionDefaultsToNone[C]](source, overrides, instances)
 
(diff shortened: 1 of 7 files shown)


ulanzetz:
This solution still provides developers to miss modifier, write recursive transformer and get stack overflow in runtime. I think .into[T] syntax should be similar to defineTo[T] to prevent runtime errors.

MateuszKubuszok:
@ulanzetz We agree that it might be the problem, however what you are proposing is a breaking change in behavior.

Since we are talking about creating implicit `Transformer[In, Out]` value, there is a high probability of shared instance e.g.:

```scala
case class User(...)
case class UserDTO(...)

object UserDTO {
  implicit val fromUser: Transformer[User, UserDTO] = ...
}

user.into[User].transform // uses UserDTO.fromUser
```

`.transformInto[T]` for a long time was just a "syntactic sugar" for `.into[T].transform` and for our users the expected behavior is that library update should not break their code. We allow ourselves to change internal implementation, but we keep the API backward compatible. By disabling scope search by default, suddenly each usage of `user.into[User].transform` requires manual intervention. What @krzemin proposed would only require change of one line - the one that defines implicit:

```scala
object UserDTO {
  implicit val fromUser: Transformer[User, UserDTO] = (_: User).defineInto[UserDTO]
}

user.into[User].transform // unchanged
```

Since I don't expect users to deploy their code untested, I would be more worried about them not understanding the cause of SO, that them seeing the error at all. For that, I imagine something like:

```scala
try {
  // run transformer
} catch {
  case e: StackOverflowException =>
    new Exception("""Transformer threw StackOverflowException
                    |You are probably defining implicit Transformer instance using .into[T].transform or .transformInto[T].
                    |Because the syntax takes the same implicit it produces recursive definition.
                    |Use .into[T].disableLocalImplicitLookup.transform or .defineInto[T] to fix this.""".stripMargin, e)
}
```

somewhere inside our library, would be enough to get them back on track.

s-nel:
Any update on this PR? Can it merge?

nadavwr:
Hey @krzemin and @MateuszKubuszok,
Thank you for this wonderful library ❤️
I tested this PR and it fixes #105 perfectly for us!
Can this please be released?

nadavwr:
Kindly reminder :)

krzemin:
Closed in favour of #129

Issue #113, opened 2019-07-09
Implicit resolution problem for `Transformer[Option[A], B]`


bjhartin:
I'm probably missing something here.

This has been discussed in #62 - mapping from options to required fields, throwing errors on `None`.  The approach there works, _if_ mapping from `Option[A]` to `A`.  However, mapping from `Option[A]` to `B` fails in some cases, _even if_ a `Transformer[A, B]` is in scope.

Example:

```scala
object ChimneyTest extends App {
  import io.scalaland.chimney.Transformer
  import io.scalaland.chimney.dsl._

  case class Bar1(i: Int) // 'extends AnyVal' makes no difference
  case class Bar2(i: String) // 'extends AnyVal' makes no difference
  case class Bar3(i: Option[Int])

  case class Baz1(b: Bar1)
  case class Baz2(b: Option[Bar1])
  case class Baz3(b: Bar2)
  case class Baz4(b: Option[Bar2])

  implicit val intToString: Transformer[Int, String] = _.toString

  implicit def idTransform[A]: Transformer[A, A] = identity[A]

  // Throws on `None` - not ideal but Chimney doesn't represent failure, e.g. `Either[Error, B]`
  implicit def transformFromOptional[A, B](implicit ev: Transformer[A, B]): Transformer[Option[A], B] =
    (a: Option[A]) => a.map(ev.transform).getOrElse(throw new NoSuchElementException("Optional field was none"))

  // (Failing example first, then simpler, passing examples)

  // FAILS TO COMPILE:
  // Baz2(Option(Bar1(1))) => Baz3(Bar2("1")) fails
  // Should be available through transformFromOptional, case class support and intToString
  //assert(Baz2(Option(Bar1(1))).transformInto[Baz3] == Baz3(Bar2("1")))

  assert(1.transformInto[String] == "1")

  assert(Option(1).transformInto[Int] == 1)

  assert(Option(1).transformInto[String] == "1")

  assert(Bar1(1).transformInto[Bar2] == Bar2("1"))

  assert(Option(Bar1(1)).transformInto[Bar1] == Bar1(1))

  assert(Baz2(Option(Bar1(1))).transformInto[Baz1] == Baz1(Bar1(1)))

  assert(Option(Bar1(1)).transformInto[Bar2] == Bar2("1"))

  assert(Baz2(Option(Bar1(1))).transformInto[Baz4] == Baz4(Option(Bar2("1"))))

[...]

PR #116, opened 2019-09-09
Add opt-in unsafe option transformation support

jchapuis:
This PR adds a `enableUnsafeOption` configuration setting, which allows for unsafe transformation from `Option[T1]` to `T2` if a transformer `T1 => T2` exists in scope.

This is really useful when working with proto3, where all fields are optional. See also issue #113.

If this can be considered for inclusion, I can work on adding coverage to the PR. We should also adapt the protobuf tests to proto3.


krzemin:
Hi @jchapuis - your proposal looks good. Having this feature under optional flag seems as reasonable move.
Please continue working on coverage and docs - we can eventually merge this.

jchapuis:
Great thanks, will continue working on it then.

jchapuis:
Ok here you go, I've added coverage and documentation, let me know what you think!

jchapuis:
Pinging you here again, am I on the right track, is there something else missing?

chimney/src/main/scala/io/scalaland/chimney/internal/TransformerMacros.scala
```diff
@@ -108,7 +110,32 @@ trait TransformerMacros {
...
+      expandTransformerTree(q"$srcPrefixTree", config.rec)(From, toInnerT).mapRight { innerTransformer =>
+        q"_root_.scala.Option[$toInnerT]($innerTransformer)"
+      }
+    }
+  }
+
+  def expandSourceWrappedInOption(srcPrefixTree: Tree, config: Config)(From: Type,
+                                                                       To: Type): Either[Seq[DerivationError], Tree] = {
+    val fromInnerT = From.typeArgs.head
+    if (From <:< noneTpe) {
```
krzemin: That’s nice. Could you please add test case that checks this behavior?

chimney/src/main/scala/io/scalaland/chimney/internal/DerivationGuards.scala
```diff
@@ -22,7 +22,11 @@ trait DerivationGuards {
   }
 
   def targetWrappedInOption(from: Type, to: Type): Boolean = {
-    to <:< optionTpe && to.typeArgs.head =:= from
+    to <:< optionTpe
+  }
+
+  def sourceWrappedInOption(from: Type, to: Type): Boolean = {
```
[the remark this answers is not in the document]
jchapuis: Yes absolutely. Good catch!

chimney/src/main/scala/io/scalaland/chimney/internal/TransformerMacros.scala
jchapuis: Sure of course

jchapuis:
Hey @krzemin I've worked on your two comments. Is there a way to compensate for the slight decrease in coverage? I think this occurs because there are more lines failing the compilation than before, right?

jchapuis:
@krzemin thanks for merging! any estimate on when we can expect a release? Thanks! 🙏

krzemin:
Tonight I preformed 0.3.3 release.

Issue #118, opened 2019-10-11
Transformation returning Validated/Either/...


sideeffffect:
Would it be possible to have a slightly different variant of `TransformerInto.enableUnsafeOption`, which instead of throwing exceptions would return a value, like `cats.data.Validated` (or something analogous, like `Either`).

Our use case is that we would like to convert ScalaPB generated case classes (where everything is `Option`al) to our domain case classes where some fields are mandatory. It would help us a lot, if the convertor method returned just a plain value (Validated, Either,...) instead of throwing and us having to catch the exceptions.

It would behave similar to [`henkan`](https://github.com/kailuowang/henkan#transform-between-case-classes-with-optional-field) which also has `validate` method to return `cats.data.Validated`.

MateuszKubuszok:
https://github.com/scalalandio/chimney/pull/146 implements what you described, but before we decide to merge and release it, it would be good if more eyes looked at it and maybe tested their use cases.

MateuszKubuszok:
Feature released in 0.5.0, available on OSS Sonatype, should appear on Maven soon.

PR #120, opened 2019-10-29
Error validation transformer (VTransformer)

ulanzetz:
This PR adds chimney-validated module with VTransformer, which can use unsafe transformations like Option unwrapping (can be usefull with proto3 generated case classes (#118)), int from string parsing etc, but unlike  enableUnsafeOption all errors are aggregated to chain and contains full path to failed fields. In my code I use cats.data.Validated, but it can be rewritten by own error-chaining semigroup implementations.

Also. now it's limited by 22 fields in single case class.

I'd be thankful for review and feedback


sideeffffect:
can I somehow help with this? I would really like to use it :smiley:

sideeffffect:
you probably need to update scoverage to 1.6.1+
https://github.com/scoverage/sbt-scoverage/issues/295

REDNBLACK:
@krzemin Can we (me and @ulanzetz) ask u kindly, to give a feedback and tell what need to be done in order for this PR to get approved? Or do u have anything against increased complexity/etc?

sideeffffect:
@REDNBLACK @ulanzetz the Travis CI build is failing for Scala 2.13. Upgrading `scoverage` might help, as I've hinted at https://github.com/scoverage/sbt-scoverage/issues/295

ulanzetz:
@sideeffffect scoverage version is already 1.6.1. Master build also failed on 2.13

sideeffffect:
maybe try changing the Travis config, so that we test on 2.13.1 (currently it's 2.13.0)

ulanzetz:
Thanks, @sideeffffect. Bumping version to 2.13.1 in travis config helped

ulanzetz:
@krzemin, what about this PR? Could you give any feedback for this. Does it make sense to continue?

krzemin:
Hi @ulanzetz, I was finally able to take a look at your PR. Below some general thoughts.

1. I think that it would be very much desired to have some form of #118 implemented in Chimney.
2. What is most problematic in the proposal, is that the whole macro is duplicated in order to support validation transformers. That would make maintenance of library much more difficult as any change or bugfix would have to be applied in two independent places. It's necessary to have derivation backbone implemented once that works for both cases and extract only crucial differences.
3. Direct dependency on Cats is troublesome in terms of being dependent on cats release cycle. It also seems wrong from the architectural standpoint: any integration with specific, external library should happen in minimal pluggable module, but the whole core implementation should not rely on any specific validation wrapper type.
4. If 2. and 3. can be addressed somehow, I would like to see such validated transformers as part of core library (along with total transformers and patchers), hence creating separate module/artifact would not be necessary.
5. Ideally other data types than cats' `Validated` (like `Option`, `Try`, any `F[_]` for which we can provide glueing code) should be supported.

Thank you for your contribution effort - I think that it totally make sense to continue working on this. I can offer some hints how to tackle these issues.

ulanzetz:
Resolved as part of https://github.com/scalalandio/chimney/pull/145

PR #125, opened 2019-12-02
Use .iterator when changing collection types.

htmldoug:
# Summary
Changes macro expansion:

| | |
| -- | -- |
| before | `Set(1,2,3).map(f).to[Seq]` |
| after | `Set(1,2,3).iterator.map(f).to[Seq]` |

## Performance
I initially spotted this as a performance problem under the profiler. Creating unnecessary `scala.collection.immutable.HashSet` instances is [expensive](http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#sets-and-maps-are-slow)!

## Correctness
There's also a potential correctness problem with mapping over `Set`, as demonstrated by the new (somewhat contrived) unit test.

/cc @weisong-rallyhealth this should speed up indexing.


krzemin:
Good catch 👍 Thanks.

PR #129, opened 2020-01-17
Transformer definition DSL

### LINKED ISSUE #105: Ability to use dsl inside Transformer definition

krzemin:
This PR provides new way to define transformer instances.

### Motivation
So far it was only possible to implement `Transformer[A, B]` type class by hand. A common use case for such definitions was to use Chimney's DSL to derive such transformation with less boilerplate. But it has few drawbacks:
- assigning an instance immediately to implicit val/dev (which is common practice) may lead to stack overflows (at runtime) or `wrong forward reference` errors (at compile time), as reported in #105 
- a programmer was forced to work with additional variable which usually was not really important from the point of transformer definition, like:

```scala
implicit val aToBtransformer: Transformer[A, B] = (aValue: A) => {
  aValue.into[B]
    .withFieldConst(...)
    .withFieldRenamed(...)
    // other modifiers
    .transform
}
```
In such examples `aValue` is not really used in other way than to begin work with Chimney DSL. Indeed it's a little additional boilerplate.

### Proposal

This PR intoruces new way of defining implicit instances:

```scala
implicit val aToBtransformer: Transformer[A, B] =
  Transformer.define[A, B]
    .withFieldConst(...)
    .withFieldRenamed(...)
    // other modifiers
    .buildTransformer
}
```

Note that:
- all DSL modifiers like (`withFieldConst`, `withFieldComputed`, etc.) are supported in definition DSL; in fact, code was reorganized in such way that their actual implementation is shared between those two DSLs
- with using `Transformer.define` we are able to specify definition scope when doing type class derivation which allows us to avoid looking up for local implicit instance and fix #105 in more compatible way
- few tests were added to verify that recursive data structures are supported (including mutually recursive data types) as long as their instances are derived and put into `implicit def`
- this PR also addresses issue with `Transformer.derive[A, B]`, as [commented here](https://github.com/scalalandio/chimney/issues/105#issuecomment-568081277)

It doesn't automatically fix problems with old style definitions (so manual intervention will be required in order to fix the code), but at least we will have one blessed way of defining transformers which I believe solve 99% of use cases in elegant way.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -66,7 +66,7 @@ lazy val root = project
   .aggregate(chimneyJVM, chimneyJS)
   .dependsOn(chimneyJVM, chimneyJS)
 
-lazy val chimney = crossProject(JSPlatform, JVMPlatform, NativePlatform)
+lazy val chimney = crossProject(JSPlatform, JVMPlatform/*, NativePlatform*/)
   .crossType(CrossType.Pure)
   .dependsOn(protos % "test->test")
   .settings(
@@ -78,13 +78,13 @@ lazy val chimney = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .settings(settings: _*)
   .settings(publishSettings: _*)
   .settings(dependencies: _*)
-  .nativeSettings(nativeLinkStubs := true)
+//  .nativeSettings(nativeLinkStubs := true)
 
 lazy val chimneyJVM = chimney.jvm
 lazy val chimneyJS = chimney.js
-lazy val chimneyNative = chimney.native
+//lazy val chimneyNative = chimney.native
 
-lazy val protos = crossProject(JSPlatform, JVMPlatform, NativePlatform)
+lazy val protos = crossProject(JSPlatform, JVMPlatform/*, NativePlatform*/)
   .crossType(CrossType.Pure)
   .settings(
     moduleName := "chimney-protos",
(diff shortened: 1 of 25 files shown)


REDNBLACK:
That’s just... Absolute madness. You actually solved it!

AviMualem:
Any estimation regarding when the PR will be merged ?

isaias-b:
I ran into this already fixed issue today and the fix was to not use the DSL and use the `Transformer.define` instead. Shouldn't this be reflected on the README.md ? I didn't even knew about this alternative way of defining custom transformations. Is the DSL deprecated by this?

krzemin:
> Is the DSL deprecated by this?

In fact, no. Transformation DSL works perfectly fine for inline, ad-hoc transformations. The problem is only when you use it in the context of defining self-referencing implicit value. For those use cases, transformer definition DSL is recommended (which beside few details has the same set of operations and semantics).

> Shouldn't this be reflected on the README.md ?

It is reflected in the [documentation](https://scalalandio.github.io/chimney/transformers/own-transformations.html#transformer-definition-dsl), isn't that enough?

isaias-b:
Never mind, i was too dumb to find that 🤦  documentation, although it is mentioned at the end of the README 🙄. 
Thanks for clearing up, that describes it appropriately! 👍

Issue #138, opened 2020-01-25
Automatic filling of Unit parameters


Krever:
It would be good if parameters of type `Unit` would be automatically filled in, as there is exactly one value of this type. 

Rationale: 
```
case class DbEntry[Id](id: Id, name: String)
type NewEntry = DbEntry[Unit]

case class NewEntryRequest(name: String)

import io.scalaland.chimney.dsl._
(_: NewEntryRequest).into[NewEntry].transform
```

krzemin:
Why not extend it for any singleton type?

Krever:
You mean literal-based singleton types (in 2.13) or (companion) object types or generally stuff like `value.type` ? It probably makes sense, I just don't use them that much and haven't thought of it.  But is there a reliable method of distinguishing singleton types?  https://www.scala-lang.org/api/current/scala/Singleton.html ?

krzemin:
Ok, technically speaking `Unit` is not singleton type from the Scala's type system perspective, despite it contains only single inhabitant. You're right to say that singleton types are not frequently used to model data types, so maybe there's no need to complicate library's code in order to fully support them.

But adding support for auto-filling target fields `f: Unit` with `f = ()` seems to be fine. 2 questions remains:
- do we need opt-out from this behavior (by `.disableAutoFillUnits`)?
- or maybe it should be turned off by default and require explicit opt-in (by `.enableAutoFillUnits`)?

Krever:
So what other ways of filling in such type would there be? You can't even use `null` as `Unit` is an `AnyVal`. The only alternative to `()` would be to throw an exception, which hardly makes any sens to me in transformer logic.

The only close-to-real danger I can imagine is to have unspotted fields of this type, coming out of some type-level shenanigans. But even if this case, chimney would just "not be a barrier" for those fields, but still won't be doing anything incorrect. 

So my vote goes definitely for "enabled by default". If opt-out is not a lot of work it could be added, but I'd rather wait for someone to request it, as I cant imagine a probable use-case for it.

ldrygala:
Hi, I would like to give it a try.

Krever:
Awesome, thank you @ldrygala  :)

PR #141, opened 2020-02-03
automatic filling of Unit parameters

ldrygala:
fixes: https://github.com/scalalandio/chimney/issues/138

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
@@ -662,8 +662,7 @@ trait TransformerMacros extends TransformerConfiguration {
           }
         }
         .orElse {
-          val targetTypeIsUnit = target.tpe <:< typeOf[Unit]
-          if (targetTypeIsUnit) {
+          if (target.tpe <:< typeOf[Unit]) {
             Some(ResolvedTargetTree(q"()"))
           } else {
             None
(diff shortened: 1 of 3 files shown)


chimney/src/test/scala/io/scalaland/chimney/DslSpec.scala
```diff
@@ -420,14 +420,14 @@ object DslSpec extends TestSuite {
     "support common data types" - {
 
       case class Foo(value: String)
-      case class Bar(value: String)
+      case class Bar(value: String, unit: Unit)
```
krzemin: I suggest to keep these test as they are. Instead, I would add new test case "support Unit" where you construct simple example with unit field in target case class.

chimney/src/test/scala/io/scalaland/chimney/DslSpec.scala
```diff
@@ -676,6 +676,14 @@ object DslSpec extends TestSuite {
 
         transform[String](polySource) ==> monoTarget
       }
+      
+      "automatically fill Unit parameters" - {
+        case class Foo(value: String)
+        case class Bar[T](value: String, poly: T)
+        type UnitBar = Bar[Unit]
+
+        Foo("test").transformInto[UnitBar] ==> Bar("test", ())
```
krzemin: What about not using type alias, but `transformInto[Bar[Unit]]` directly?

chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
```diff
@@ -661,6 +661,14 @@ trait TransformerMacros extends TransformerConfiguration {
             None
           }
         }
+        .orElse {
+          val targetTypeIsUnit = target.tpe <:< typeOf[Unit]
+          if (targetTypeIsUnit) {
```
krzemin: seems like this type test could be inlined

ldrygala:
thanks @krzemin for helping me with my first contribution 👌

PR #142, opened 2020-02-03
missing test for automatic Unit parameter filling

ldrygala:
missing test for: #138


chimney/src/test/scala/io/scalaland/chimney/DslSpec.scala
```diff
@@ -441,8 +441,10 @@ object DslSpec extends TestSuite {
       "support automatically filling of scala.Unit" - {
         case class Buzz(value: String)
         case class NewBuzz(value: String, unit: Unit)
+        case class FooBuzz(unit:Unit)
```
krzemin: Shouldn't field be named `value: Unit` to test name-conflicting behavior?

ldrygala: good point 👍

PR #145, opened 2020-02-11
TransformerF and cats.data.Validated support

ulanzetz:
Add TransformerF[F[_], A, B] for transformation A to B in F context.
Transformer[A, B] is TransformerF for identity F now.
Backward compatibility is not broken, all DSL logic was saved.
Add support for user-defined contexts of transformation and context based on cats.data.Validated for error aggretated transformations in separate module, it can be useful for unwrapping options for proto3 generated case classes.

This is better version of https://github.com/scalalandio/chimney/pull/120 without copypaste


chimney-cats/src/main/scala/io/scalaland/chimney/cats/validated/package.scala
```diff
@@ -0,0 +1,33 @@
...
+import io.scalaland.chimney.{TransformationContext, TransformationException}
+
+package object validated {
+  type NEC[+A] = cats.data.NonEmptyChain[A]
+  val NEC = cats.data.NonEmptyChain
+
+  type V[+A] = cats.data.Validated[NEC[TransformationError], A]
+  val V = cats.data.Validated
+
+  implicit val vContextError: TransformationContext[V] = new TransformationContext[V] {
```
krzemin: Do you think providing instance for any applicative functor would be possible?

chimney/src/main/scala/io/scalaland/chimney/TransformationContext.scala
```diff
@@ -0,0 +1,19 @@
...
+  def map[A, B](fa: F[A])(f: A => B): F[B]
+
+  def product[A, B](fa: F[A], fb: F[B]): F[(A, B)]
+
+  def addPrefix[A](fa: F[A], prefix: String): F[A]
+
+  def error[A](message: String): F[A]
+}
+
+object TransformationContext {
```
krzemin: FYI: I provided built-in instances for `Option` and `Either[M[E], *]` here.

chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
```diff
@@ -81,58 +83,67 @@ trait TransformerMacros extends TransformerConfiguration {
...
+          expandTargetWrappedInOption(srcPrefixTree, config, ctx)(From, To, F)
+        } else if (isOption(From)) {
+          expandSourceWrappedInOption(srcPrefixTree, config, ctx)(From, To, F)
+        } else if (bothEithers(From, To) && isId(F)) {
           expandEithers(srcPrefixTree, config)(From, To)
-        } else if (bothMaps(From, To)) {
+        } else if (bothMaps(From, To) && isId(F)) {
           expandMaps(srcPrefixTree, config)(From, To)
-        } else if (bothOfTraversableOrArray(From, To)) {
+        } else if (bothOfTraversableOrArray(From, To) && isId(F)) {
```
krzemin: It should be possible to generate collections (also maps, eithers, etc.) transformers for any `F`, but you would need `traverse` in `TransformerContext`.

chimney-cats/src/main/scala/io/scalaland/chimney/cats/validated/package.scala
ulanzetz: I think it's possible to provide instance for any applicative error functor, but it can be fast-fail and it's not nice for such transformations

chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
ulanzetz: Yes, I was thinking about this, but it's non-optimal to do this for any iterable.
I want to safe bothOfTraversableOrArray for backward compatibility in id context and require Traverse[G] instance in other contexts.

ulanzetz: @krzemin, I implemented Traverse for cats.Traverse collections and fixed all old test cases. The only place where compatibility was broken is TransformationException instead of NoSuchElementException in enableUnsafeOptions. Also I think it's should be support of maps in F context

chimney-cats/src/main/scala/io/scalaland/chimney/cats/validated/TransformationError.scala
```diff
@@ -0,0 +1,21 @@
+package io.scalaland.chimney.cats.validated
+
+abstract class TransformationError(val info: String) {
+  def addPrefix(prefix: String): TransformationError
+}
+
+object TransformationError {
+  private case class Deep(message: String, path: NEC[String])
+      extends TransformationError(s"$message on ${path.toChain.iterator.mkString(".")}") {
```
Phill101: Could you provide access to the `path` from `TransformationError`?
It would be nice to have a possibility to interact directly with it.

Moreover, it seems like you reconstruct this message every time when you adding a prefix.
Correct me if I'm wrong.

So how about to define an `info` as `def` instead of `val` and to provide access to the `path`?

krzemin:
Thank you for your efforts @ulanzetz!

I've created #153 in order to explore enhanced error reporting for lifted transformers, which was proposed here.

PR #146, opened 2020-02-18
Lifted transformers

krzemin:
This PR is implementation of #118 approached in a bit different way to alternative #145 proposed by @ulanzetz.

Differences highlight:
1. unlike to #145, this one keeps pure `Transformer` type class, along with `TransformerF`; the macro is generalized to be able to derive any transformer (pure or wrapped) as the derivation structure and rules seems to be the same; #145 removes `Transformer` and uses `TransformerF[Id, From, To]` for pure transformers
2. in both PRs there is supporting type class that need to be defined any wrapper type `F[_]`; in this PR there is  [`TransformerFSupport`](https://github.com/scalalandio/chimney/compare/feature/lifted-transformers?expand=1#diff-6a1b5c35c5b01bf3a6cc9b87acf9bfa1), in #145 there is [`TransformationContext`](https://github.com/scalalandio/chimney/pull/145/files#diff-27bc60e364c3fdcbf06cc6f9bcbeb679); there are minor differences in signatures (and naming), but they are mostly do the same
3. due to rewrite of mapping algorithm in this PR, we are able to mix pure values (`withFieldConst`/`withFieldComputed`) with wrapped ones (`withFieldConstF`/`withFieldComputedF`) in such a way that pure ones doesn't need to be boxed using `TransformerFSupport.pure`; [as far as I understand](https://github.com/scalalandio/chimney/pull/145/files#diff-9b882292dbf2a36fc5c11b7ec17ed514R111), #145 always wraps in `TransformationContext.pure`, even in case `F = Id`
4. #145 provides [enhanced support for capturing more detailed transformation errors](https://github.com/scalalandio/chimney/pull/145/files#diff-688b0a9036ef614af1122c6aa172edf9) with object path hold in a context; this PR doesn't support it, but it may be added later (I need to look closer at #145 proposal)
5. #145 provides [a module for cats Validated integration](https://github.com/scalalandio/chimney/pull/145/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR100), while in this PR there are instances only for `Option[*]` and `Either[M[E], *]`

After a brief look, both solutions should provide source compatibility, however in this PR several compile error messages may have changed due to refactor of mapping algorithm.

TODO list includes:
- adding more tests, verifying border cases more comprehensively
- completing implementation of wrapped transformers derivation for cases like options, eithers, sealed families, etc.
- further refactor of transformer configuration (both type level in DSL and compile-time value level)
- completing DSL operations in `TransformerFDefinition` and `TransformerFInto`

I believe we can pick best of both (this and #145) to the final release.

Any feedback is highly appreciated.


chimney/src/main/scala/io/scalaland/chimney/dsl/TransformerInto.scala
```diff
@@ -20,6 +21,9 @@ final class TransformerInto[From, To, C <: TransformerCfg](
     val td: TransformerDefinition[From, To, C]
 ) {
 
+  def lift[F[_]]: TransformerFInto[F, From, To, WrapperType[F, C]] =
+    new TransformerFInto[F, From, To, WrapperType[F, C]](source, td.lift[F])
```
ulanzetz: Why no implicit TransformerFSupport[F] here and in another places where F[_] appears?

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
```diff
@@ -0,0 +1,97 @@
...
+  implicit def TransformerFEitherErrorAccumulatingSupport[E, C[X] <: IterableOnce[X]](
+      implicit ef: Factory[E, C[E]]
+  ): TransformerFSupport[Either[C[E], *]] = new TransformerFSupport[Either[C[E], *]] {
+
+    def pure[A](value: A): Either[C[E], A] = Right(value)
+
+    def combine[A, B](fa: Either[C[E], A], fb: => Either[C[E], B]): Either[C[E], (A, B)] = {
+      (fa, fb) match {
+        case (Right(ok1), Right(ok2)) =>
+          Right((ok1, ok2))
```
ulanzetz: Is it necessary to support each IterableOnce? Maybe use one type of collection with const concat time

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
```diff
@@ -0,0 +1,97 @@
...
+
+trait TransformerFSupport[F[_]] {
+
+  def pure[A](value: A): F[A]
+
+  def combine[A, B](fa: F[A], fb: => F[B]): F[(A, B)]
+
+  def map[A, B](fa: F[A], f: A => B): F[B]
+
+  def traverse[M[X] <: Iterable[X], A, B](mfa: M[A], f: A => F[B])(
```
Phill101: Iterable upper bound restricts users to use collections which implement this interface.
It makes impossible to use third-party collections like `cats.data.Chain`.
Isn't it? If you use type class as a constraint, I think it will be much better.

What do you think?

chimney/src/main/scala/io/scalaland/chimney/dsl/TransformerInto.scala
krzemin: Implicit instance of `F[_]` is resolved by a macro that expands transformer (where it's really needed in order to construct transformer tree). We could add the instance here, but it would be not used. We couldn't capture it on the type-level configuration (as the resolved instance is a value-level thing).

Theoretically we could carry it on the value-level (as a member of `TransformerFDefinition`), but such member would have to be public and it wouldn't be worth the effort.

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
```diff
@@ -0,0 +1,97 @@
...
+
+object TransformerFSupport {
+
+  implicit def TransformerFOptionSupport: TransformerFSupport[Option] =
+    new TransformerFSupport[Option] {
+
+      def pure[A](value: A): Option[A] = Option(value)
+
+      def combine[A, B](fa: Option[A], fb: => Option[B]): Option[(A, B)] = {
+        fa.flatMap(a => fb.map(b => (a, b)))
```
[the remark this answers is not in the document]
krzemin: Unfortunately, it doesn't work with Scala 2.11/2.12:
```
[error] TransformerFSupport.scala:27:15: type mismatch;
[error]  found   : Iterable[(A, B)]
[error]  required: Option[(A, B)]
[error]         fa.zip(fb)
[error]               ^
[info] Iterable[(A, B)] <: Option[(A, B)]?
[info] false
```

It seems that `.zip` on `Option` was added in 2.13 during collections rework. In 2.11/2.12 it was available through optionToIterable implicit conversion, but resulting type is Iterable - hence the type error.

It could be fixed by using `.headOption` that compiles fine on all 2.11-2.13:
```scala
fa.zip(fb).headOption
```

But I'm not sure whether it's good idea as there is additional iterator materialization under the hood.

Alternative possible implementation which is more like `.zip` from 2.13:

```scala
def combine[A, B](fa: Option[A], fb: => Option[B]): Option[(A, B)] = {
  lazy val fbv = fb
  if (fa.isEmpty || fbv.isEmpty) None else Some((fa.get, fbv.get))
}
```

chimney/src/main/scala/io/scalaland/chimney/dsl/TransformerDefinition.scala
```diff
@@ -18,6 +19,18 @@ final class TransformerDefinition[From, To, C <: TransformerCfg](
     val instances: Map[(String, String), Any]
 ) {
 
+  def __refineConfig[C1 <: TransformerCfg]: TransformerDefinition[From, To, C1] =
+    this.asInstanceOf[TransformerDefinition[From, To, C1]]
+
+  def __addOverride(key: String, value: Any): TransformerDefinition[From, To, C] =
+    new TransformerDefinition(overrides.updated(key, value), instances)
+
```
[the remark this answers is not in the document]
krzemin: In `TransformerDefinition` there is not much advantage beside being consistent with `TransformerFDefinition`.

In `TransformerFDefinition` it's a longer story:
- these methods allow to overcome some difficulties with accessing to type tree of `F[_]`
- in case of macro, we can request `WeakTypeTag` instance for type parameters, but it doesn't really work with types with arguments (it requires kind `*`, we need to wildcard type args)
- materialization of such type tree ended up with not picking the right implicit instance of `TransformerFSupport` during tree materialization (seems that type constructed in this way was somehow broken)
- using `_addOverride` and `__refineConfig` allowed to eliminate access to `F` in a macro - it's just passed through by these operations

Unfortunately, they must be public as they are invoked by the macro-expanded code. But so far we have few public stuff in `Transformer(F)Definition/Into`. It would be nice to get rid of them, but one must propose new encoding that eliminates them.

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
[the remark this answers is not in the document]
Phill101: > it doesn't work with Scala 2.11/2.12

I didn't know that.

```scala
fa.flatMap(a => fb.map(b => (a, b)))
```
much better I think 😄

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
krzemin: Let's make a distinction between `F[_]` and `M[_]` here.

`traverse` is invoked for every field of type `Seq[A]` that needs to be transformed into `Seq[B]`, but we have (or can derive) `TransformerF` between `A` and `B`, so a function of type `A => F[B]`; `traverse` allows us to lift the function on a collection level, so we can obtain `Seq[A] => F[Seq[B]]`. In user's code we will not have only `Seq`s, but probably other types. The minimal interface that we need in order to provide reasonable implementation of `traverse` is `IterableOnce`.

I can guess why you are worried about `cats.data.Chain` (correct me if I got it bad) - you would like to capture transformation errors into `cats.data.Chain`, for example: in `ValidatedNec[T]` (which Chain, or NonEmptyChain is a part of) which is an instantiation of `F[_]`, not `M[_]`.

`M[_]` here expresses all the types that automatic conversions are supported by Chimney out of the box (which is Iterable in fact). If you want to support any other `M[_]` which is not `Iterable`, which may happed if you use `cats.data.Chain` to model your domain, you just need to provide  implicit instance of `TransformerF[F[_], M[A], M[B]]`, so actually: `M[A] => F[M[B]]`, probably given an instance of `TransformerF[F[_], A, B]`, so `A => F[B]`.

If you want, I may provide a test case that illustrates it.

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
krzemin: Well, in core library this is the widest type `C` that we can provide implementation for `Either[C[E], A]`. This is one instance only and no one stops you from providing instance for more specialized types. In cats integration module we can still define `TransformerFSupport[ValidatedNec]` without requiring `IterableOnce` for error type.

As in Scala there's no standard type class/bound for expressing collection with constant concat time, and I'm not keen to provide one in Chimney, I guess that's fine, isn't it?

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
Phill101: You are right. I confused with `F[_]`

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
ulanzetz: Maybe use own ADT instead of Either with UnrolledBuffer or another collection optimal for concatenation in internal and case error sequence to List or another immutable collection in public api?

chimney/src/main/scala/io/scalaland/chimney/dsl/TransformerDefinition.scala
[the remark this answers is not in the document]
ulanzetz: Why not use single TransformerDefinition for TransformerF and Trasformer (with F[_] = Id)?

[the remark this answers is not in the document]
krzemin: This could reduce code volume (and I will try to do this later), but I'm afraid will not address original issue with WeakTypeTags for type constructors.

I tried to look at your implementation and you may hit the same issue, because of [this](https://github.com/scalalandio/chimney/pull/145/files#diff-28b933308957ae0c4ae2f692ccf28782R92). Just try to play a bit with `Either[Seq[E], *]`.

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
krzemin: Currently I don't think that it's Chimney's duty to provide default collection types for optimal error concatenation. In many use cases standard `Vector` (or even `List`) will be just fine. For others, where performance does matter:
- it will either work ootb (as long as the collection type inherits IterableOnce) - hence I think `Either[UnrolledBuffer[String], *]` should just work now,
- or user will just need to provide few lines of glue code to make anything he needs working (such an instance combining Either and IterableOnce might be a good example how to implement it for your own type).

I agree that we should underline performance implications in the documentation for lifted transformers.

chimney/src/main/scala/io/scalaland/chimney/TransformerF.scala
```diff
@@ -0,0 +1,21 @@
+package io.scalaland.chimney
+
+import io.scalaland.chimney.dsl.TransformerFDefinition
+import io.scalaland.chimney.internal.TransformerCfg
+import io.scalaland.chimney.internal.macros.ChimneyBlackboxMacros
+
+import scala.language.experimental.macros
+
+trait TransformerF[F[_], From, To] {
```
MateuszKubuszok: Will need docs at some point.

chimney/src/main/scala/io/scalaland/chimney/Transformer.scala
```diff
@@ -20,7 +20,7 @@ object Transformer {
   /** Provides [[io.scalaland.chimney.Transformer]] derived with the default settings.
     *
     * @tparam From type of input value
-    * @tparam Into type of output value
+    * @tparam To type of output value
```
MateuszKubuszok: Good catch.

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
```diff
@@ -0,0 +1,93 @@
+package io.scalaland.chimney
+
+import scala.collection.compat._
+
+trait TransformerFSupport[F[_]] {
```
MateuszKubuszok: Will need docs.

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
```diff
@@ -0,0 +1,93 @@
...
+
+  def pure[A](value: A): F[A]
+
+  def combine[A, B](fa: F[A], fb: => F[B]): F[(A, B)]
+
+  def map[A, B](fa: F[A], f: A => B): F[B]
+
+  def traverse[M, A, B](it: Iterator[A], f: A => F[B])(
```
MateuszKubuszok: In particular we need to explain that here we use `Iterator` only **internally** at it doesn't impose limitations of what structures can be handled by derivation.

chimney/src/main/scala/io/scalaland/chimney/dsl/TransformerDefinition.scala
```diff
@@ -18,6 +18,18 @@ final class TransformerDefinition[From, To, C <: TransformerCfg](
     val instances: Map[(String, String), Any]
 ) {
 
+  def __refineConfig[C1 <: TransformerCfg]: TransformerDefinition[From, To, C1] =
```
MateuszKubuszok: Add comment that these are public only because macro requires it to be public.

Actually, we could probably rewrite it to be extension method of some sort but stored in companion object, so that it would be always available and had access to private fields (and name these extension methods something like "InternalSth")

chimney/src/main/scala/io/scalaland/chimney/dsl/TransformerFDefinition.scala
```diff
@@ -0,0 +1,34 @@
...
+import io.scalaland.chimney.internal.macros.{ChimneyBlackboxMacros, TransformerDefinitionWhiteboxMacros}
+
+import scala.language.experimental.macros
+
+final class TransformerFDefinition[F[_], From, To, C <: TransformerCfg](
+    val overrides: Map[String, Any],
+    val instances: Map[(String, String), Any]
+) {
+
+  def __refineConfig[C1 <: TransformerCfg]: TransformerFDefinition[F, From, To, C1] =
```
MateuszKubuszok: Same case as in TransformerDefinition

chimney/src/main/scala/io/scalaland/chimney/internal/macros/DslWhiteboxMacros.scala
```diff
@@ -1,95 +1,109 @@
...
-    val C = weakTypeOf[C]
-
-    val fieldNameConst = Constant(fieldName.decodedName.toString)
-    val fieldNameLit = Literal(fieldNameConst)
-    val singletonFieldTpe = c.internal.constantType(fieldNameConst)
-    val newCfgTpe = tq"_root_.io.scalaland.chimney.internal.TransformerCfg.FieldConst[$singletonFieldTpe, $C]"
-    val fn = TermName(c.freshName("ti"))
+    q"""
+      ${c.prefix.tree}
+        .__addOverride(${fieldName.toNameLiteral}, $value)
```
MateuszKubuszok: How about  some extension methods allowing something like:
```scala
c.prefix.tree
  .addOverride(fieldName.toNameLiteral, value)
  .refineConfigFieldConst(fieldName.toSingletonTpe, weakTypeOf[C]) // maybe sth better?
```
it would be easier to refactor in the future if any update/rename/change could be less stringly.

I mean here and in the following quasiquites.

krzemin: If you mean extension methods on trees (not adding this stuff to public api), I'm fine with such refactor 👍

chimney/src/main/scala/io/scalaland/chimney/dsl/TransformerDefinition.scala
krzemin: Such extension methods would be still visible in IDE code completion, so benefit would be minimal (if any).

But the trick with hiding these methods in some implicit class (which is not automatically imported) could be possible. The macro will need to ensure that it's imported, when needed; it requires prefix inspection (to avoid importing it several times), but this way we could hide these methods from appearing in IDE.

I will give it a try later.

krzemin: In fact, implicit part is not absolutely crucial to this idea. We can have extension class (`TransformerDefinitionOps`) without implicits, as we drive the code generation here. Instead of `td.__refineConfig[C]` we just emit `TransformerDefinitionOps(td).refineConfig[C]`.

This way, we still don't need to put `F` type tree to emitted code, while remove these internal `__` methods from public API.

krzemin: I've tried this. It works well in 2.12/2.13, but due to some problems with type constructors inference, it fails in 2.11. It applies both to first idea (with implicit class + import) and to second (plain class + new in generated code).

As for now, I think that keeping these methods in public api (until we drop 2.11) is lesser evil. I will add proper comments though.

chimney/src/main/scala/io/scalaland/chimney/dsl/TransformerFDefinition.scala
krzemin: added

chimney/src/main/scala/io/scalaland/chimney/TransformerF.scala
krzemin: docs added

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
krzemin: docs added

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
krzemin: explained

chimney/src/main/scala/io/scalaland/chimney/internal/macros/DslWhiteboxMacros.scala
krzemin: done ✅

krzemin:
Now I think almost all of the bullets from TODO list was addressed. There is comprehensive [test suite](https://github.com/scalalandio/chimney/pull/146/files#diff-1588cd20fece8ba31da0cca9f4a51e63) which should cover all the cases in the context of lifted transformers. 

I think it's a good moment to take another round of review. cc: @ulanzetz, @Phill101, @MateuszKubuszok 

Next steps would be:
- [ ] cats Validated module
- [ ] decide what to do with enhanced error reporting
- [ ] write docs

chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import _root_.cats.data.Validated
+import _root_.cats.kernel.Semigroup
+
+import scala.collection.compat._
+
+package object cats {
+
+  implicit def TransformerFValidatedSupport[EE: Semigroup]: TransformerFSupport[Validated[EE, +*]] =
+    new TransformerFSupport[Validated[EE, +*]] {
```
ulanzetz: Maybe add cats.data.Ior for semigroup A instance too?

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
```diff
@@ -0,0 +1,154 @@
...
+      }
+    }
+
+    def traverse[M, A, B](it: Iterator[A], f: A => Either[C[E], B])(
+        implicit fac: Factory[B, M]
+    ): Either[C[E], M] = {
+      val bs = fac.newBuilder
+      val eb = ef.newBuilder
+
+      while (it.hasNext) {
```
ldrygala: maybe it would make sense to introduce some flag (similar to `wasNone` from Option) to avoid collecting rights when there already exist left

chimney/src/main/scala/io/scalaland/chimney/internal/utils/EitherUtils.scala
```diff
@@ -18,12 +18,21 @@ trait EitherUtils {
       }
     }
 
+    def flatMapRight[A1 >: A, B1](f: B => Either[A1, B1]): Either[A1, B1] = {
+      either match {
```
ldrygala: why not `Either.flatMap`?

chimney/src/main/scala/io/scalaland/chimney/internal/utils/EitherUtils.scala
```diff
@@ -32,8 +41,31 @@ trait EitherUtils {
...
           value
         case _ =>
+          // $COVERAGE-OFF$
           throw new NoSuchElementException
+        // $COVERAGE-ON$
+      }
+    }
+
+    def rightOrElse[A1 >: A, B1 >: B](or: => Either[A1, B1]): Either[A1, B1] = {
+      either match {
```
ldrygala: why not `Either.orElse` ?

chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
krzemin: Looks like a good idea 👍 

But I'm struggling with reasonable semantics of
```scala
def product[A, B](fa: Ior[EE, A], fb: => Ior[EE, B]): Ior[EE, (A, B)]
```
for few cases:
- when `fa = Left(e)` and `fb = Right(b)` - we cannot have `Ior.Both`, as we can't construct right pair (`(A, B)`), since we only have `B`; `Ior.Left` is the only result we can have here
- when `fa = Left(e1)` and `fb = Both(e2, b)` - we can combine `e1` and `e2` using Semigroup instance, but also we cannot have `Ior.Both` as a result
It also includes symmetrical cases.

I'm not really familiar with using `Ior`, but I'm afraid that `product` can't be implemented with substantial information loss. The question is if such semantics make sense at all?

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
```diff
@@ -0,0 +1,154 @@
...
+        case (left1, left2) =>
+          val eb = ef.newBuilder
+          left1.left.foreach(eb ++= _)
+          left2.left.foreach(eb ++= _)
+          Left(eb.result())
+      }
+    }
+
+    def map[A, B](fa: Either[C[E], A], f: A => B): Either[C[E], B] = {
+      fa match {
```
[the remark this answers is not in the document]
krzemin: `Either.map` was introduced in Scala 2.12. As long as we support 2.11, we can't unfortunately :(

chimney/src/main/scala/io/scalaland/chimney/internal/utils/EitherUtils.scala
krzemin: Not present in Scala 2.11

chimney/src/main/scala/io/scalaland/chimney/internal/utils/EitherUtils.scala
krzemin: Not available in 2.11

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
krzemin: good idea 👍 also, when first error appears, we can clear `bs` builder immediately

krzemin: ✅ done

chimney/src/main/scala/io/scalaland/chimney/dsl/package.scala
```diff
@@ -8,35 +8,67 @@ import scala.language.experimental.macros
...
+      * @return transformed value of target type `To`
       */
     final def transformInto[To](implicit transformer: Transformer[From, To]): To =
       transformer.transform(source)
   }
 
+  implicit class TransformerFOps[From](private val source: From) extends AnyVal {
+
+    /**
+      * Allows to customize wrapped transformer generation to your target type.
```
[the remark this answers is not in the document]
krzemin: yes

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
krzemin: Note about performance of picked collection type added to docs.

krzemin:
As this PR is already huge enough, I think that enhanced error path reporting (as in @ulanzetz's #145 proposal), although highly welcomed, might be added later as a separate PR. Doing it right is a major challenge itself and I don't intend to make review of this PR even harder for anyone.

chimney/src/main/scala/io/scalaland/chimney/dsl/TransformerDefinition.scala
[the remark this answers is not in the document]
krzemin: > Why not use single TransformerDefinition for TransformerF and Trasformer (with F[_] = Id)?

Lifted transformer (`TransformerF`) is conceptually different thing to total transformer (`Transformer`). We could possibly define `Transformer[A, B] = TransformerF[Id, A, B]` and not have duplication in DSLs. But having them separated:
- we can emit efficient code without wrapping in `.pure` for `Id`
- we can mix lifted and partial DSL operations, lifting in place when first used `withFieldConstF`/`withFieldComputedF`/etc.
- keeping current `Transformer` untouched will not force current users to rewrite their code

It seems as a reasonable tradeoff to keep these types separated. At some point later we may revisit this belief.

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
```diff
@@ -0,0 +1,160 @@
+package io.scalaland.chimney
+
+import scala.collection.compat._
+
+/** Type class supporting lifted transformers.
+  *
+  * In order to create lifted transformation from `A` to `F[B]`,
+  * we need these few operations to be implemented for specific `F`
+  * wrapper type.
```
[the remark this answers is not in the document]
krzemin: Isn't this explained enough in a scaladoc of `traverse`?

Krever:
I cant go through code right now but I tried to think about cases in which this feature interacts with my quite long experience with using chimney in production. 
Weirdly, I never felt the need for lifted transformers or at least I can't recall such need. 
Although I remember that chimney had an `unsafe` option to transform `Option[T] => T` via `.get` which now may stop being unsafe.  I can't check right now if this was addressed somehow in the PR, so I'm just leaving the comment for your consideration. 

Everything else looks good! I'm quite astonished by the amount of work put in this. Great job!

krzemin:
@Krever this PR actually keeps  `Option[T] => T` as unsafe even in the lifted context, but it provides `withFieldConstF` which allows you to express your domain validation rules, possibly capturing validation errors. I think that in the long term `Option[T] => T` can be marked at deprecated in favour of lifted transformers (which is already recommended in docs as a safer alternative).

ulanzetz:
@Krever also you can define something like this
```
 implicit def optionUnwrapped[F[_]: ApplicativeError, A, B](implicit underlying: Transformer[A, B]): TransformerF[F, Option[A], B] =
    {
      case Some(value) => ApplicativeError[F].pure(underlying.transform(value))
      case None            => ApplicativeError[F].raiseError(...)
    } 
```
and make your option unwrapping safe

PR #148, opened 2020-03-07
Disallow method by default searching source type

### LINKED ISSUE #128: Don't use fields or methods by default?
```
object Test {

  case class Boo(i: Int, s: String) {
    def d: Double = 5
  }

  case class Foo(i: Int, s: String, d: Double)

  Boo(1, "").into[Foo].transform

}
```

This compiles on version 0.3.5, but doesn't feel like it should. Should we only consider case class constructor fields for automatic field mapping?

jatcwang:
- Add enableMethodCalls setting to allow methods with no parameter list

Resolves #128

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/dsl/TransformerDefinition.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/dsl/TransformerDefinition.scala
@@ -70,6 +70,16 @@ final class TransformerDefinition[From, To, C <: TransformerCfg](
   def enableUnsafeOption: TransformerDefinition[From, To, EnableUnsafeOption[C]] =
     this.asInstanceOf[TransformerDefinition[From, To, EnableUnsafeOption[C]]]
 
+  /** Enable values to be supplied from method calls. Source method must be public and have no parameter list.
+    *
+    * By default this is disabled because method calls may perform side effects (e.g. mutations)
+    *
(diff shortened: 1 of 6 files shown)


chimney/src/main/scala/io/scalaland/chimney/dsl/TransformerDefinition.scala
```diff
@@ -70,6 +70,16 @@ final class TransformerDefinition[From, To, C <: TransformerCfg](
...
     this.asInstanceOf[TransformerDefinition[From, To, EnableUnsafeOption[C]]]
 
+  /** Enable values to be supplied from method calls. Source method must be public and have no parameter list.
+    *
+    * By default this is disabled because method calls may perform side effects (e.g. mutations)
+    *
+    * @see [[https://scalalandio.github.io/chimney/#UsingMethodCalls]] for more details
+    * @return [[io.scalaland.chimney.dsl.TransformerDefinition]]
+    */
+  def enableMethodCalls: TransformerDefinition[From, To, EnableMethodCalls[C]] =
```
krzemin: Suggest changing name to `enableMethodAccessors`. "Calls" part is a bit misleading in my opinion.

jatcwang: Yep sounds good :) Done

PR #150, opened 2020-03-07
Properly resolve generic type signatures for fields in PatcherMacros

### LINKED ISSUE #149: Patching using higher kinded data
I recently hit the limitation of patching API when I tried to use it in combination with higher kinded data. Please see examples below:

```scala
case class EntryT[Id](id: Id)
type Entry = EntryT[EntryId]
case class Patch(id: EntryId)

(x: Entry).patchUsing((u: Patch))
```

```scala
case class Data[F[_]](name: F[String])
case class Real(name: String)
type Patch = Data[Option]

(x: Real).patchUsing(y: Patch)
```
In both cases the errors are similar:

```
derivation from patch.id: Id to Entry.Id is not supported in Chimney!
```

```
  derivation from patch.name: Data.F to java.lang.String is not supported in Chimney!
```

Could the params be substituted for their actual value being used? I think similar works for the transofmers.

krzemin:
It resolves #149.


Krever:
I published locally and it solves the mentioned issue indeed.

PR #154, opened 2020-03-19
WIP Error path support

### LINKED ISSUE #153: Enhanced error reporting for lifted transformers
Currently in lifted transformers user is responsible for creating invalid case of `F[To]`, which most likely ends up like:

```scala
.withFieldComputedF(_.field, source => parse(source.field) match {
  case valid =>
    // construct F[To] based on valid
    Validation.valid(valid)
  case invalid =>
    // pack <invalid> case to F[To], with any eventual error info from `source`
    Validation.invalid(NEC.one(s"${source.id}: invalid!"))
})
```

Despite having `source` available in context brings a possibility to provide more detailed error information, this approach has some disadvantages:
1. wrapping into `F` needs to be done manually each time (`Validation.invalid(NEC.one(<error message comes here>))`), while only error message is relevant here
2. it does not provide detailed context information (i.e. path to concrete object in the source entity) in case this transformation is re-used in outer one

In #145 there was prototype implementation of enhanced error reporting which addressed (2).

We need to explore possibilities on how to enhance API in a compatible way in order to address both issues.

ulanzetz:
Partially resolve #153.

Add path for case classes and iterable/array, but I suppose it need also correctly works with maps.
For maps support we should change Map -> Map expanding rule.
What do you think, @krzemin?


chimneyCats/src/main/scala/io/scalaland/chimney/cats/TransformationError.scala
```diff
@@ -0,0 +1,31 @@
+package io.scalaland.chimney.cats
+
+import cats.data.{Chain, NonEmptyChain}
+
+abstract class TransformationError(val message: String) {
```
krzemin: 1. Is it necessary that this type is abstract? It seems like it could be a case class containing a message and an error path (possibly empty). WDYT?

2. Is it possible to abstract over concrete error message type? Currently in lifted transformers we may just use strings, as good as some custom type that describe error.

chimneyCats/src/test/scala/io/scalaland/chimney/cats/CatsValidatedSpec.scala
```diff
@@ -442,6 +442,52 @@ object CatsValidatedSpec extends TestSuite {
           .intoF[ValidatedNec[String, +*], long.NumScale[Int]]
           .transform ==> Validated.invalidNec("bad int")
       }
+
+      "path of error" - {
+        type V[+A] = ValidatedNec[TransformationError, A]
+
+        def printError(err: TransformationError): String =
```
krzemin: Maybe it's good candidate as `def describe: String`  on `TransformationError`?

chimney/src/main/scala/io/scalaland/chimney/TransformerFErrorPathSupport.scala
```diff
@@ -0,0 +1,5 @@
+package io.scalaland.chimney
+
+trait TransformerFErrorPathSupport[F[+_]] {
```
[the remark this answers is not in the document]
ulanzetz: If we want to extend it from TransformerFSupport and implement for example instance for Either[C[TransformationError], +*], we need to make TransformerFEitherErrorAccumulatingSupport as class and extends it in TransformerFErrorPathSupport implementation.

Also we need to do it for Validated instance

I think it's not necessary to extend, users do something like this
```
implicit val instance: TransformerFSupport[V] with TransformerFErrorPathSupport[V] = { ... } 
```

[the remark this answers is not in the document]
krzemin: Oh, yes - sure!

But what I meant is that maybe we don't need to define `TransformerFErrorPathSupport` and add `addPath` method directly to `TransformerFSupport`?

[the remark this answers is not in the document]
ulanzetz: How implement it for option and Either[C[E], +*] instances? Just return fa? I think it's not good idea

chimneyCats/src/main/scala/io/scalaland/chimney/cats/TransformationError.scala
ulanzetz: Made message type abstract, made case class instead of adt

chimneyCats/src/test/scala/io/scalaland/chimney/cats/CatsValidatedSpec.scala
ulanzetz: For abstract message type?

chimneyCats/src/main/scala/io/scalaland/chimney/cats/TransformationError.scala
```diff
@@ -0,0 +1,31 @@
...
+  def errorPath: Chain[String]
```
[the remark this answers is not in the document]
ulanzetz: Moved to core module, add instance for Either[C[TransformationError[M], +*]]

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
```diff
@@ -65,6 +65,12 @@ trait TransformerFSupport[F[+_]] {
     */
   def traverse[M, A, B](it: Iterator[A], f: A => F[B])(
       implicit fac: Factory[B, M]
+  ): F[M] =
+    traverseWithIndex[M, A, B](it, (a, _) => f(a))
+
+  // TODO: add doc
+  def traverseWithIndex[M, A, B](it: Iterator[A], f: (A, Int) => F[B])(
```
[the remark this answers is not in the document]
ulanzetz: We can add counter in our macro outside of traverse function, but it will fail, if traverse will implemented parallel for example

chimneyCats/src/main/scala/io/scalaland/chimney/cats/TransformationError.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  def addPrefix(prefix: String): TransformationError
```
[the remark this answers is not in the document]
ulanzetz: Added ADT

chimney/src/main/scala/io/scalaland/chimney/TransformerFErrorPathSupport.scala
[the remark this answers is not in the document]
krzemin: I agree, it doesn't make sense. Let's keep the separate type class.

In fact, it allows us to even implement error prefixing support for concrete `E`s. It means that one may provide implementation for `Either[C[String], +*]` and have all of the usual strings prefixed with error path - that's nice 👍

chimneyCats/src/test/scala/io/scalaland/chimney/cats/CatsValidatedSpec.scala
krzemin: Nope, doesn't make sense any more :)

chimney/src/main/scala/io/scalaland/chimney/TransformationError.scala
```diff
@@ -0,0 +1,6 @@
+package io.scalaland.chimney
+
+case class TransformationError[M](message: M, errorPath: List[ErrorPathNode] = Nil) {
+  def prepend(node: ErrorPathNode): TransformationError[M] =
+    TransformationError[M](message, errorPath.prepended(node))
+}
```
krzemin: Can we add some method that returns nicely formatted string (like: `"foo.bar[5].baz"`)?

chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala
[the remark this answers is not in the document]
krzemin: What if we pass an iterator, which is already zipped with index?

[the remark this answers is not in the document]
ulanzetz: Nice idea! Fixed

chimney/src/main/scala/io/scalaland/chimney/TransformationError.scala
ulanzetz: Added

krzemin:
This PR really looks promising! ❤️ 

Regarding Map rule, we're now able to treat map like iterable and generate basically the same code as for maps, lists, seqs, etc. That means that in case of enhanced error reporting, we obtain `Index` path element with some int (a result of `map.iterator.zipWithIndex`) which is rather useless.
For maps I believe we should have separate case for `ErrorPathNode` which directly represents map keys.

Other than that, we should check if other rules (like eithers, optionals, tuples, java beans) require some special treatment from the perspective of this feature. We should have at least some tests for each one.

And finally, write documentation (both scaladocs for new typeclass and user manual).

ulanzetz:
@krzemin I added maps support, review please

chimney/src/main/scala/io/scalaland/chimney/ErrorPathNode.scala
```diff
@@ -0,0 +1,33 @@
...
+    val separator: String = "."
+  }
+
+  case class Index(value: Int) extends ErrorPathNode {
+    def show: String = s"[$value]"
+
+    val separator: String = ""
+  }
+
+  case class MapKey[K](key: K) extends ErrorPathNode {
```
krzemin: Does `K` carry any reasonable type-level information when being element of `List[ErrorPathNode]`?
It seems it will be LUB-bed to AnyRef anyway, so I guess `key: AnyRef` may be sufficient here.

chimney/src/main/scala/io/scalaland/chimney/ErrorPathNode.scala
```diff
@@ -0,0 +1,33 @@
+package io.scalaland.chimney
+
+sealed trait ErrorPathNode {
+  def show: String
+
+  def separator: String
+}
+
+object ErrorPathNode {
+  case class Accessor(name: String) extends ErrorPathNode {
```
krzemin: All of those can be `final`.

chimney/src/main/scala/io/scalaland/chimney/ErrorPathNode.scala
```diff
@@ -0,0 +1,33 @@
...
+    def show: String = name
+
+    val separator: String = "."
```
krzemin: Suggest making it `def`, as it will be materialized eagerly even if one doesn't invoke pretty-printing paths.

Alternatively, instead of both fields you may define separate functions in companion object.

chimneyCats/src/test/scala/io/scalaland/chimney/cats/CatsValidatedSpec.scala
```diff
@@ -442,6 +442,72 @@ object CatsValidatedSpec extends TestSuite {
...
+
+        "map" - {
+          case class Foo(map: Map[String, String], map2: Map[String, String])
+
+          case class Bar(map: Map[Int, Int], map2: Map[String, Int])
+
+          Foo(Map("a" -> "b", "c" -> "d"), Map("e" -> "f")).transformIntoF[V, Bar].leftMap(_.map(printError)) ==>
+            Validated.Invalid(
+              NonEmptyChain(
+                "Can't parse int from a on map.keys",
```
krzemin: I curious about few more cases:

1. What would be the path in case that we have case class as map keys? I believe for `case class CCKey(value: String)` it should be printed as `map[CCKey(a)]`, shouldn't it?

2. I see that you implemented path reporting for case where map key can't be transformed. That's very nice 👍 But will it always end on `map.keys`? What if we have some advanced transformation in map keys that also carry some path info? Would it be composed somehow?

chimney/src/test/scala/io/scalaland/chimney/DslFSpec.scala
```diff
@@ -786,5 +786,98 @@ object DslFSpec extends TestSuite {
           .transform ==> Right(OuterOut(InnerOut("pure: test")))
       }
     }
+
+    "error path" - {
```
krzemin: This file is huge enough already. WDYT about creating separate test suite for testing error path reporting for lifted transformers?

chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
```diff
@@ -129,6 +132,8 @@ trait TransformerMacros extends TransformerConfiguration with MappingMacros with
       expandSourceWrappedInOption(srcPrefixTree, config)(From, To)
     } else if (bothEithers(From, To)) {
       expandEithers(srcPrefixTree, config)(From, To)
+    } else if (bothMaps(From, To)) {
```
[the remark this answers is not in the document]
ulanzetz: Generalized rule for case when only `From` is map. We can't handle it in `expandIterableOrArray` cause resolveRecursiveTransformerBody for tuple (key, value) and it will be tagged by tuple rule

chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
```diff
@@ -609,13 +687,28 @@ trait TransformerMacros extends TransformerConfiguration with MappingMacros with
...
     )(
       accessor.symbol.resultTypeIn(From),
       target.tpe
     )
+
+    config.wrapperErrorPathSupportInstance match {
+      case Some(errorPathSupport) =>
+        resolved.mapRight { bodyTree =>
+          if (bodyTree.isWrapped)
+            bodyTree.copy(tree = q"""$errorPathSupport.addPath[${target.tpe}](
```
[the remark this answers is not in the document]
ulanzetz: I'm not sure it's better, it is used in 4 places with different path node types and extension method should receive lot of trees and types

chimney/src/main/scala/io/scalaland/chimney/ErrorPathNode.scala
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  case class Index(value: Int) extends ErrorPathNode {
+    def show: String = s"[$value]"
```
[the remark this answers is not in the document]
ulanzetz: Changed to scala-like

chimney/src/main/scala/io/scalaland/chimney/ErrorPathNode.scala
ulanzetz: Fixed

chimney/src/main/scala/io/scalaland/chimney/ErrorPathNode.scala
ulanzetz: Fixed

chimney/src/main/scala/io/scalaland/chimney/ErrorPathNode.scala
ulanzetz: Changed to def

chimneyCats/src/test/scala/io/scalaland/chimney/cats/CatsValidatedSpec.scala
ulanzetz: 1. Added test
2. Value of key is able in transformer and directly adding to error message, for example in my case message `"Can't parse int from a on map.keys"` contains information in what case it failed

chimney/src/test/scala/io/scalaland/chimney/DslFSpec.scala
ulanzetz: Moved to `ErrorPathSpec` and `CatsValidatedErrorPathSpec`

chimneyCats/src/test/scala/io/scalaland/chimney/cats/CatsValidatedSpec.scala
krzemin: 1. 👍 
2. The information comes from a message provided by user in error case of function passed to `withFieldComputedF`, not from a path itself. The path here informs us that the problem relates to some key, but we don't know which one (just by looking at the path). Let me elaborate on more advanced example:

```scala
case class KeyFoo(innerField: String)
case class KeyBar(innerField: Int)

case class Foo(map: Map[KeyFoo, Boolean])
case class Bar(map: Map[KeyBar, Boolean])
```

Having implicit `TransformerF[..., String, Int]` and trying to transform:

```scala
Foo(Map(KeyFoo("nan") -> True)).transformIntoF[..., Bar]
```

...my question was about what error path we obtain in this case?

Note that `map.keys` path doesn't really inform us which key the error regards. From the inner transformation of `KeyFoo` to `KeyBar` we can extract path information so that it failed because `innerField` field can't be parsed to int.

Do you think we can somehow enrich the path so that it carry both information?
In this case it should be something similar to `map.KeyFoo(nan).innerField` (I'm not stubborn about the syntax).

ulanzetz: Fixed

chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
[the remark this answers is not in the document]
krzemin: Ok. Current implementation delegating all other cases to `expandIterableOrArray` seems fine too.

chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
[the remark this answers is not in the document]
krzemin: Ok, let's leave it for now and revisit in final review again.

ulanzetz:
Added docs, @krzemin, I think all features of error path are ready, but I think we should also add safe option unwrapping to next release

MateuszKubuszok:
Over the weekend, I'll look at this one last time and ping @krzemin to move on with this PR.

ulanzetz:
@krzemin what about this? Maybe something should be changed? Is it able to merge?

ncreep:
Hi,

Given that #183 was merged, I'm very eager for this pull request to be merged as well (dreaming of nice error-reporting when dealing with Protobuf...).

Is there anything I can do to help speed this along? Fix the conflicts? Any further required changes to the pull request?

Thanks

krzemin (changes requested):
I've made another round of review, left few remarks.

chimney/src/test/scala/io/scalaland/chimney/ErrorPathSpec.scala
```diff
@@ -0,0 +1,144 @@
+package io.scalaland.chimney
+
+import io.scalaland.chimney.dsl._
+import io.scalaland.chimney.internal.utils.EitherUtils._
+import io.scalaland.chimney.utils.OptionUtils._
+
+import utest._
+
+object ErrorPathSpec extends TestSuite {
+  val tests = Tests {
```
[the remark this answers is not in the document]
ulanzetz: Added error path node for override fields and test for withFieldConstT and withFieldComputedT usage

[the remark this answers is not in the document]
krzemin: Hmm, this is not what I meant. I don't think that we need to treat overrides in any special way. The error path `foo.overrides(bar)` seems to be kind of misleading and I'm not really sure if I understand what does it mean. Would you mind reverting that?

What I had in mind was to add the test case that uses [transformer definition DSL](https://scalalandio.github.io/chimney/transformers/own-transformations.html#transformer-definition-dsl) for lifted transformers, so that:
```scala
Transformer
   .defineF[..., ...]
   ...
   .buildTransformer
```

which interacts with error paths support is tested for correct behavior.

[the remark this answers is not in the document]
ulanzetz: Reverted overrides error path capturing. Updated test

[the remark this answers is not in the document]
krzemin: Great, thank you 👍

krzemin (approved):
I'm still not yet sure about all design decisions that have been made here.
However this implementation curated for a long time, has pretty good tests coverage and I believe it's better to merge it now as release as experimental feature, allowing for potential breaking changes in case we discover some issues later.

Thank you @ulanzetz for your time and determination 👍

PR #155, opened 2020-03-23
Safe option unwrapping support

ulanzetz:
While writing docs for #154, I discovered that one of the main case for error path - option unwrapping doesn't work. And I can't define something like this
```
implicit def optUnwrap[A, B](implicit underlying: Transforner[A, B]): TransformerF[V, A, B] = ???
```
Only if explicitly define it for each pairs of types.

I think we should add type class provides auto safe option unwrapping.
Maybe union it with error path support type class and think about name, something like `TransformerFErrorSupport`

TransformationError in this PR can merge TransformationError in #154 and obtain path.

 WDYT, @krzemin?

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/TransformationError.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/TransformationError.scala
@@ -1,0 +1,9 @@
+package io.scalaland.chimney
+
+sealed trait TransformationError[+M]
+
+object TransformationError {
+  case object OptionUnwrappingError extends TransformationError[Nothing]
+
+  case class WithMessage[+M](message: M) extends TransformationError[M]
+}
--- a/chimney/src/main/scala/io/scalaland/chimney/TransformerFFromOptionSupport.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/TransformerFFromOptionSupport.scala
@@ -1,0 +1,35 @@
+package io.scalaland.chimney
+
+import scala.collection.compat._
+
+trait TransformerFFromOptionSupport[F[+_]] {
+
+  /** Create F[B] from Option[A]
+    *
+    * @tparam A type of wrapped value
+    * @param opt wrapped in option value
+    * @param trans wrapped transformation function from A to B in F
+    * @return wrapped in F result of transformation option unwrapping
+    */
+  def fromOption[A, B](opt: Option[A], trans: A => F[B]): F[B]
+}
+
+object TransformerFFromOptionSupport {
+  implicit val OptionInstance: TransformerFFromOptionSupport[Option] = new TransformerFFromOptionSupport[Option] {
+    override def fromOption[A, B](opt: Option[A], trans: A => Option[B]): Option[B] =
+      opt.flatMap(trans)
+  }
+
+  implicit def EitherStringInstance[C[X] <: IterableOnce[X], M](
+      implicit ef: Factory[TransformationError[M], C[TransformationError[M]]]
+  ): TransformerFFromOptionSupport[Either[C[TransformationError[M]], +*]] =
+    new TransformerFFromOptionSupport[Either[C[TransformationError[M]], +*]] {
+      override def fromOption[A, B](
+          opt: Option[A],
+          trans: A => Either[C[TransformationError[M]], B]
+      ): Either[C[TransformationError[M]], B] =
+        opt.fold[Either[C[TransformationError[M]], B]](
+          Left(ef.fromSpecific(List(TransformationError.OptionUnwrappingError)))
+        )(trans)
+    }
+}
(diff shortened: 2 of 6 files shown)


krzemin:
I was thinking about deprecating unsafe option unwrapping (`.enableUnsafeOption`) which always throws an exception, even in the lifted context, so I think we shouldn't bother too much.

We cannot really implement `Option[A] -> B` safely using total transformers, but suggested replacement should be to use lifted ones. It is possible even now, but one must manually handle each field which is not really convenient (especially in proto3 context, where every field is optional). So I think it's good to think about more automated replacement for safe option unwrapping 👍 

I briefly looked at your proposal and I'm not very happy to introduce another type class. I understand we cannot handle this case generically as long as we want to support generic error types for Either/Validated. But I think the problem is deeper and we are missing some abstraction. (I also think this is the reason why we need separate type class for error path prepend in #154).

I have brief proposal that tries to address the problem.

What if we are stubborn and want to add following operations to `TransformerFSupport[F[+_]]`:
```scala
  def prependErrorPath[A](fa: F[A], errorPathNode: ErrorPathNode): F[A]
  def raiseEmptyOption: F[A]
```

Can we implement both these operations for `F = Option`? Of course!

```scala
  def prependErrorPath[A](fa: F[A], errorPathNode: ErrorPathNode): F[A] = Option.empty[A] // just ignore the path, as Option can't hold it
  def raiseEmptyOption: F[A] = Option.empty[A]
```

Can we implement them for `F = Either[C[E], +_]`? It depends on E.

Let's assume `E = String`.
```scala
  def prependErrorPath[A](fa: F[A], errorPathNode: ErrorPathNode): F[A] =
    fa.left.map { errors => errors.map(e => s"${errorPathNode.separator}${errorPathNode.show}: $e") }
  def raiseEmptyOption: F[A] =
    Left(C("Empty option value!"))
    // pseudo code, in fact we use factory to construct C collection
```

What if `E = TransformationError[E1]`, but this one defined in #154 ?
```scala
  def prependErrorPath[A](fa: F[A], errorPathNode: ErrorPathNode): F[A] =
    fa.left.map { errors => errors.map(_.prepend(errorPathNode) }
  def raiseEmptyOption: F[A] =
    Left(C(TransformationError(???)))
    // it seems that we need to first know how to raise empty option on `E1`
```

What if `F = Validated[EE, +*]`?
- in order to implement `prependErrorPath`, we would need to require that beside `EE` being `Semigroup`, it must be a `Functor` as well, so in fact `EE = C[E]`, but also we need to know how to prepend a single error path node in `E`
- in order to implement `raiseEmptyOption` we need to be able to insert `E` to `C[E]` (`C` being `InvariantMonoidal`?), but also first how to construct `E` which signals empty option

Looking at these know and unknowns, we can define a type class:
```scala
trait ErrorTypeSupport[E] {
  def addPrefixPath(e: E, node: ErrorPrefixNode): E
  def raiseEmptyOption: E
}
```

And require an instance for given E only for those `F`s, where `E` is left abstract. I think this should eliminate need for `TransformerFErrorPathSupport` from #154, `TransformerFFromOptionSupport` from this PR, while still allowing user to define custom type for representing transformation error.

This idea probably requires some lifting in details, but generally I believe it's better for maintenance to have few more operations in `TransformerFSupport`, perhaps some of them delegated to other type class than have multiple type classes which must be resolved in macro.

Sorry for a bit lengthy reasoning, but I'm curious what do you think @ulanzetz?

ulanzetz:
If wiil we add `ErrorTypeSupport` 
```
trait ErrorTypeSupport[E] {
  def addPrefixPath(e: E, node: ErrorPrefixNode): E
  def raiseEmptyOption: E
}
```
And will require type class instance for `TransformerFSupport` instances for `Either[C[E]], +*]` and `Validated[EE[E], +*]`, it will break backward compatibility, for example `Either[List[Int], +*]` will no longer be supported. Would not it be better to add separate type class,
```
trait TransformerErrorSupport[F[+_]] {
  def prependErrorPath[A](fa: F[A], errorPathNode: ErrorPathNode): F[A]
  def raiseEmptyOption: F[A]
}
```
implement it for Option, Either[C[String], +*], Either[C[TransformationError[E], +*], etc, and keep backward compatible?

ghostdogpr:
I just found this PR and #154 which are exactly what we need. We have some large protobuf objects returning `Option` everywhere, and we'd like to replace our legacy macros with chimney, only we need the error to tell us which field was missing. I believe this and #154 would address that pain and allow us to migrate entirely to chimney.

Hope this could get merged soon 😄

nitayk:
Hello,
We are also interested in this feature - is there any way we can help push it forward?

Thanks

ulanzetz:
After merge of #183 will be available to add implicit-based rule. 
https://github.com/scalalandio/chimney/pull/183/files#diff-1588cd20fece8ba31da0cca9f4a51e63

Issue #158, opened 2020-04-03
Unable to compile when transforming Option[Seq[T] to Seq[T]


allantl:
Consider the following example:

```scala
 object Test {

  implicit def optionSeqToSeq[T, U](implicit tf: Transformer[T, U]): Transformer[Option[Seq[T]], Seq[U]] =
    new Transformer[Option[Seq[T]], Seq[U]] {
      override def transform(src: Option[Seq[T]]) = {
        src.getOrElse(Seq.empty).map(_.transformInto[U])
      }
    }

  case class ClassA(a: Option[Seq[String]])
  case class ClassB(a: Seq[String])

  def transform(a: ClassA): ClassB = {
    a.transformInto[ClassB]
  }
}
```

This results in a compilation error
```scala
[error] scala.collection.Seq
[error]   derivation from classa.a: scala.Option to scala.collection.Seq is not supported in Chimney!
```

If I provide an explicit
```scala
implicit val strToStrTransformer: Transformer[String, String] = Transformer.derive
```
Then, chimney would be able to resolve the issue. 
It would be nice if chimney can derive the required implicit automatically.

Context: I need this for proto3 conversion. I have a legacy case class that has a lot of `Option[Seq[T]]`. Since, proto only supports `Seq[T]`, I need to transform `Option[Seq[T]]` to `Seq[T]`.

ulanzetz:
@allantl you can't define optionSeqToSeq, but you can use marco-level defined transformation, using enableOptCollectionFlatting, watch my PR

allantl:
@ulanzetz Oh wow, that was quick 👍

krzemin:
I digged deeper into that issue. It seems that what's happening when chimney tries to generate transformer between `ClassA` and `ClassB` is:
1. It finds pair of fields that matches by name: `a: Option[Seq[String]]` in ClassA,  `a: Seq[String]` in ClassB
2. Before trying to recursively generate transformer, it performs implicit lookup searching for implicit instance of `Transformer[Option[Seq[String]], Seq[String]]`
3. `implicit def optionSeqToSeq[T, U](implicit tf: Transformer[T, U]): Transformer[Option[Seq[T]], Seq[U]]` is considered as a candidate, but for some reason it's rejected
4. Chimney tries to recursively derive transformation, but none of the rules is able to to it
5. Compilation fails with aforementioned error message.

[...]

PR #159, opened 2020-04-03
Resolve issue #158

ulanzetz:
Add config setting `enableOptCollectionFlattening` to allow transformation from Option[Collection[A]] to Collection[B] if transformation from A to B is defined

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/dsl/ConfigDsl.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/dsl/ConfigDsl.scala
@@ -21,8 +21,8 @@ trait ConfigDsl[CC[_ <: TransformerCfg], C <: TransformerCfg] {
     *
     * @see [[https://scalalandio.github.io/chimney/transformers/customizing-transformers.html#using-method-accessors]] for more details
     */
-  def enableOptCollectionFlatting: CC[EnableOptCollectionFlatting[C]] =
-    this.asInstanceOf[CC[EnableOptCollectionFlatting[C]]]
+  def enableOptCollectionFlattening: CC[EnableOptCollectionFlattening[C]] =
+    this.asInstanceOf[CC[EnableOptCollectionFlattening[C]]]
 
   /** Fail derivation if `From` type is missing field even if `To` has default value for it.
     *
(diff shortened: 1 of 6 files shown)


chimney/src/main/scala/io/scalaland/chimney/dsl/ConfigDsl.scala
```diff
@@ -14,6 +14,16 @@ trait ConfigDsl[CC[_ <: TransformerCfg], C <: TransformerCfg] {
...
     this.asInstanceOf[CC[EnableMethodAccessors[C]]]
 
+  /** Enable transformation from Option[Collection[A]] to Collection[B] if transformation from A to B can be derived.
+    * If Option is empty, it will be empty collection
+    *
+    * By default in such case compilation fails.
+    *
+    * @see [[https://scalalandio.github.io/chimney/transformers/customizing-transformers.html#using-method-accessors]] for more details
+    */
+  def enableOptCollectionFlatting: CC[EnableOptCollectionFlatting[C]] =
```
paulpdaniels, line 24: Nit: Shouldn't it be `Flattening`?

ulanzetz: fixed

MateuszKubuszok:
Thanks for contribution!

To be honest, I would prefer to investigate why `implicit def transformFlatten[A]: Transform[F[G[A], H[A]]` is being ignored. Adding special case in config is a last resort solution, as someone else might do something similar for other collection and have the same issue as #158. So I would prefer to figure out a root cause before we implement special-case workaround.

ulanzetz:
@MateuszKubuszok 
It's being ignored due type parameters. If we define something like this:
```implicit def dependent[A, B](implicit underlying: Transformer[A, B]): Tranformer[...]```,
it will be not derived, if Transformer[A, B] is not defined explicitly (not derived in macro).
Or we should explicitly define `dependent` for each (A, B) pair

```
implicit val dependentStringString: Transformer[...] = dependent[String, String]
implicit val dependentIntInt: Transformer[...] = dependent[Int, Int]
```

For same reason we can't define unsafe option unwrapper without marco
```
implicit def optUnwrapped[A, B](implicit underlying: Transformer[A, B]): Transformer[Option[A], B] =
    a => underlying(a.get)
```
This will also be ignored.

The only thing we can do, to add new branch in macro derviation

Issue #156, opened 2020-03-24
Suggest using .enableMethodAccessors in the error message when def accessor found


flicken:
Full minimal example: https://github.com/flicken/chimney-issue

```
object internal {
  case class Event(venue: Venue)

  sealed trait Venue {
    def name: String
  }

  case class ManuallyFilled(name: String) extends Venue
}

object dto {
 case class Event(venue: Venue)
 case class Venue(name: String)
}

 import io.scalaland.chimney.dsl._
val venue = internal.ManuallyFilled("Venue Name")
val event = internal.Event(venue)

// Compiles on both
venue.transformInto[dto.Venue]
// Fails to compile on 0.5.0, but does on 0.4.2
event.venue.transformInto[dto.Venue]
```

Failure message on 0.5.0:
```
[error] .../chimney-issue/fails/src/main/scala/net/flicken/ChimneyExample.scala:16:35: Chimney can't derive transformation from net.flicken.internal.Venue to net.flicken.dto.Venue
[error] 
[error] net.flicken.dto.Venue
[error]   name: java.lang.String - no accessor named name in source type net.flicken.internal.Venue
[error] 
[error] Consult https://scalalandio.github.io/chimney for usage examples.
[error] 
[error]     println(venue.into[dto.Venue].transform)
[error]                                   ^
[error] one error found
[error] (fails / Compile / compileIncremental) Compilation failed
```

krzemin:
That's a bug indeed.

It seems that derivation fails when upcasted to sealed base type.

```scala
(venue: internal.Venue).transformInto[dto.Venue]
```

This also fails to compile (with the same error message), but it should work.

Thanks for reporting @flicken.

krzemin:
Sorry, I misjudged the example that you provided.

It seems that this behavior is correct according to #128 / #148.

Please use turn your `def name` into `val` or `lazy val`, or enable using method accessors, like:

```scala
event.venue.into[dto.Venue].enableMethodAccessors.transform
```

Read more at https://scalalandio.github.io/chimney/transformers/customizing-transformers.html#using-method-accessors

flicken:
Ah, thanks, I missed that in the release notes.

The compilation error message could be more informative then.  I would expect something like

```
[error] net.flicken.dto.Venue
[error]   name: java.lang.String - no accessor named name in source type net.flicken.internal.Venue, but there was a method with this name
[error]
[...]

PR #160, opened 2020-04-03
add suggestion for method accessor

ldrygala:
fixes: https://github.com/scalalandio/chimney/issues/156

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/DerivationError.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/DerivationError.scala
@@ -9,7 +9,8 @@ final case class MissingAccessor(
     fieldName: String,
     fieldTypeName: String,
     sourceTypeName: String,
-    targetTypeName: String
+    targetTypeName: String,
+    defAvailable: Boolean = false
 ) extends DerivationError
 
 final case class MissingJavaBeanSetterParam(
@@ -51,8 +52,10 @@ object DerivationError {
       .map {
         case (targetTypeName, errs) =>
           val errStrings = errs.distinct.map {
-            case MissingAccessor(fieldName, fieldTypeName, sourceTypeName, _) =>
-              s"  $fieldName: $fieldTypeName - no accessor named $fieldName in source type $sourceTypeName"
+            case MissingAccessor(fieldName, fieldTypeName, sourceTypeName, _, defAvailable) =>
+              val defAvailableHint =
+                if (defAvailable) ", but there was a method with this name. Try to `.enableMethodAccessors`" else ""
+              s"  $fieldName: $fieldTypeName - no accessor named $fieldName in source type $sourceTypeName" + defAvailableHint
             case MissingJavaBeanSetterParam(setterName, requiredTypeName, sourceTypeName, _) =>
               s"  set${setterName.capitalize}($setterName: $requiredTypeName) - no accessor named $setterName in source type $sourceTypeName"
             case MissingTransformer(fieldName, sourceFieldTypeName, targetFieldTypeName, sourceTypeName, _) =>
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/DerivationError.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/DerivationError.scala
@@ -44,18 +44,15 @@ final case class NotSupportedDerivation(fieldName: String, sourceTypeName: Strin
     extends DerivationError
 
 object DerivationError {
-
   def printErrors(errors: Seq[DerivationError]): String = {
 
     errors
-      .groupBy(_.targetTypeName)
+      .groupBy(e => (e.targetTypeName, e.sourceTypeName))
       .map {
-        case (targetTypeName, errs) =>
+        case ((targetTypeName, sourceTypeName), errs) =>
           val errStrings = errs.distinct.map {
-            case MissingAccessor(fieldName, fieldTypeName, sourceTypeName, _, defAvailable) =>
-              val defAvailableHint =
-                if (defAvailable) ", but there was a method with this name. Try to `.enableMethodAccessors`" else ""
-              s"  $fieldName: $fieldTypeName - no accessor named $fieldName in source type $sourceTypeName" + defAvailableHint
+            case MissingAccessor(fieldName, fieldTypeName, sourceTypeName, _, _) =>
+              s"  $fieldName: $fieldTypeName - no accessor named $fieldName in source type $sourceTypeName"
             case MissingJavaBeanSetterParam(setterName, requiredTypeName, sourceTypeName, _) =>
               s"  set${setterName.capitalize}($setterName: $requiredTypeName) - no accessor named $setterName in source type $sourceTypeName"
             case MissingTransformer(fieldName, sourceFieldTypeName, targetFieldTypeName, sourceTypeName, _) =>
@@ -70,8 +67,22 @@ object DerivationError {
               s"  derivation from $fieldName: $sourceTypeName to $targetTypeName is not supported in Chimney!"
           }
 
+          val fieldsWithMethodAccessor = errors.collect {
+            case MissingAccessor(fieldName, _, _, _, true) => s"`$fieldName`"
+          }
+          val methodAccessorHint =
+            if (fieldsWithMethodAccessor.nonEmpty) {
+              val first3Fields = fieldsWithMethodAccessor.take(3).mkString(", ")
+              val otherFields = fieldsWithMethodAccessor.length - 3
+              val fields = if (otherFields > 0) s"$first3Fields and $otherFields other method" else first3Fields
+
+              s"There are methods in $sourceTypeName that might be used as accessors for $fields fields in $targetTypeName. Consider using `.enableMethodAccessors`"
+            } else ""
+
           s"""$targetTypeName
            |${errStrings.mkString("\n")}
+           |
+           |$methodAccessorHint
            |""".stripMargin
       }
       .mkString("\n")
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/DerivationError.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/DerivationError.scala
@@ -76,12 +76,11 @@ object DerivationError {
               val otherFields = fieldsWithMethodAccessor.length - 3
               val fields = if (otherFields > 0) s"$first3Fields and $otherFields other methods" else first3Fields
 
-              s"There are methods in $sourceTypeName that might be used as accessors for $fields fields in $targetTypeName. Consider using `.enableMethodAccessors`"
+              s"\nThere are methods in $sourceTypeName that might be used as accessors for $fields fields in $targetTypeName. Consider using `.enableMethodAccessors`"
             } else ""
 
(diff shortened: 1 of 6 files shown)


chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
```diff
@@ -587,14 +584,16 @@ trait TransformerMacros extends TransformerConfiguration with MappingMacros with
...
-            )
+          accessorsMapping(target) match {
+            case Accessor.Resolved(symbol: MethodSymbol, wasRenamed: Boolean) =>
+              MissingTransformer(
+                fieldName = target.name,
+                sourceFieldTypeName = symbol.resultTypeIn(From).typeSymbol.fullName,
+                targetFieldTypeName = target.tpe.typeSymbol.fullName,
+                sourceTypeName = From.typeSymbol.fullName,
+                targetTypeName = To.typeSymbol.fullName
+              ) +: erroredTargets(target)
```
MateuszKubuszok, line 595: The order of reporting errors changed, before we were appending but now we are prepending. I think we could leave the previous behavior (appending) just in case.

chimney/src/test/scala/io/scalaland/chimney/DslSpec.scala
```diff
@@ -596,7 +596,7 @@ object DslSpec extends TestSuite {
       "method is disabled by default" - {
         compileError("""Foobar("param").into[Foobar3].transform""").check(
           "",
-          "method: java.lang.String - no accessor named method in source type io.scalaland.chimney.DslSpec.Foobar"
+          "method: java.lang.String - no accessor named method in source type io.scalaland.chimney.DslSpec.Foobar, but there was a method with this name. Try to `.enableMethodAccessors`"
```
MateuszKubuszok, line 599: Now that I think about it, there migh be more cases when we would provide some hints to the user. In that case should we always provide a separate ADT just to trace things? Perhaps, we could have one common metadata, for storing these, which would be extensible? After all we use it only if compilation fails to generate more precise error message?

Not necessarily an issue for now, but I would like to discuss this for the future.

chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
```diff
@@ -575,7 +572,7 @@ trait TransformerMacros extends TransformerConfiguration with MappingMacros with
       Right(resolvedBodyTrees)
     } else {
       val targetsToFallback = erroredTargets.collect {
-        case (target, _) if accessorsMapping(target).isEmpty =>
+        case (target, _) if !accessorsMapping(target).isInstanceOf[Accessor.Resolved] =>
```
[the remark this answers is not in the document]
ldrygala: 👍 fixed

chimney/src/main/scala/io/scalaland/chimney/internal/macros/Model.scala
```diff
@@ -20,6 +20,16 @@ trait Model extends TransformerConfiguration {
   }
 
   case class TransformerBodyTree(tree: Tree, isWrapped: Boolean)
-  case class ResolvedAccessor(symbol: MethodSymbol, wasRenamed: Boolean)
 
+  sealed trait Accessor extends Product with Serializable
```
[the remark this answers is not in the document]
ldrygala: 👍 fixed

chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
ldrygala: 👍 fixed

chimney/src/main/scala/io/scalaland/chimney/internal/macros/MappingMacros.scala
```diff
@@ -145,16 +152,27 @@ trait MappingMacros extends Model with TransformerConfiguration {
...
     val sourceName = ms.name.decodedName.toString
     if (config.enableBeanGetters) {
       val lookupNameCapitalized = lookupName.capitalize
-      sourceName == lookupName ||
-      sourceName == s"get$lookupNameCapitalized" ||
-      (sourceName == s"is$lookupNameCapitalized" && ms.resultTypeIn(From) == typeOf[Boolean])
+      if (sourceName == lookupName ||
+          sourceName == s"get$lookupNameCapitalized" ||
+          (sourceName == s"is$lookupNameCapitalized" && ms.resultTypeIn(From) == typeOf[Boolean])) {
+        AccessorResolution.Resolved(ms, wasRenamed)
```
krzemin, line 162: We don't yet support renaming java beans getters. This change is not sufficient in order to provide such support. And I think we can safely set this flag to `false` to better communicate the intent.

chimney/src/main/scala/io/scalaland/chimney/internal/macros/Model.scala
```diff
@@ -20,6 +20,19 @@ trait Model extends TransformerConfiguration {
...
 
   case class TransformerBodyTree(tree: Tree, isWrapped: Boolean)
-  case class ResolvedAccessor(symbol: MethodSymbol, wasRenamed: Boolean)
 
+  sealed trait AccessorResolution extends Product with Serializable {
+    def isResolved: Boolean
+  }
+  object AccessorResolution {
+    case object NotFound extends AccessorResolution {
+      override val isResolved: Boolean = false
```
krzemin, line 29: Isn't it better to keep it (and other `isResolved`) as `def`?

chimney/src/main/scala/io/scalaland/chimney/internal/macros/MappingMacros.scala
```diff
@@ -54,8 +54,15 @@ trait MappingMacros extends Model with TransformerConfiguration {
         }
         val wasRenamed = lookupName != target.name
         fromGetters
-          .find(lookupAccessor(config, lookupName, wasRenamed, From))
-          .map(ms => ResolvedAccessor(ms, wasRenamed))
+          .map(lookupAccessor(config, lookupName, wasRenamed, From))
+          .sortBy {
```
krzemin, line 58: Why sorting is needed here? Can't we simplify it to finding first resolution that is other than `NotFound`? Like:

```scala
fromGetters
  .find(ms => lookupAccessor(config, lookupName, wasRenamed, From)(ms) != AccessorResolution.NotFound)
  .getOrElse(AccessorResolution.NotFound)
```

chimney/src/main/scala/io/scalaland/chimney/internal/macros/Model.scala
ldrygala: I think it's quite normal to define something as `def` in trait and override it using `val`. I don't see any advantage of using `def` here

chimney/src/main/scala/io/scalaland/chimney/internal/macros/MappingMacros.scala
ldrygala: my intention was to take `AccessorResolution.Resolved` over `AccessorResolution.DefAvailable`.  Not sure if it's possible to have `val` and method which fit to some target field

chimney/src/main/scala/io/scalaland/chimney/internal/macros/Model.scala
krzemin: `val` causes immediate memory allocation, even if it's not called. The way that we use in `resolveAccessorsMapping` suggests that for common use case a lot of `NotFound` objects are produced, for which `isResolved` has no chance to be called.

Not a big thing, but I think making it `def` would make it more GC friendly.

chimney/src/main/scala/io/scalaland/chimney/internal/macros/MappingMacros.scala
krzemin: My thesis is that you can't have both `Resolved` and `DefAvailable` for given `fromGetter` list. If that's true, you can safely eliminate sorting. But I might be wrong. The easiest way to be sure is to add a test that ensures this.

chimney/src/main/scala/io/scalaland/chimney/internal/macros/Model.scala
ldrygala: 👌

chimney/src/main/scala/io/scalaland/chimney/internal/macros/MappingMacros.scala
ldrygala: 👌

chimney/src/main/scala/io/scalaland/chimney/internal/DerivationError.scala
```diff
@@ -51,8 +52,10 @@ object DerivationError {
       .map {
         case (targetTypeName, errs) =>
           val errStrings = errs.distinct.map {
-            case MissingAccessor(fieldName, fieldTypeName, sourceTypeName, _) =>
-              s"  $fieldName: $fieldTypeName - no accessor named $fieldName in source type $sourceTypeName"
+            case MissingAccessor(fieldName, fieldTypeName, sourceTypeName, _, defAvailable) =>
+              val defAvailableHint =
+                if (defAvailable) ", but there was a method with this name. Try to `.enableMethodAccessors`" else ""
```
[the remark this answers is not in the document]
ldrygala: fixed

chimney/src/main/scala/io/scalaland/chimney/internal/macros/MappingMacros.scala
```diff
@@ -137,24 +144,36 @@ trait MappingMacros extends Model with TransformerConfiguration {
         }
 
       defaultValueFallback orElse optionNoneFallback orElse unitFallback
-    }.toMap
+    }
```
krzemin, line 147: `.to(ListMap)` doesn't work?

chimney/src/main/scala/io/scalaland/chimney/internal/DerivationError.scala
```diff
@@ -67,8 +67,22 @@ object DerivationError {
...
+              val first3Fields = fieldsWithMethodAccessor.take(3).mkString(", ")
+              val otherFields = fieldsWithMethodAccessor.length - 3
+              val fields = if (otherFields > 0) s"$first3Fields and $otherFields other methods" else first3Fields
+
+              s"There are methods in $sourceTypeName that might be used as accessors for $fields fields in $targetTypeName. Consider using `.enableMethodAccessors`"
+            } else ""
+
           s"""$targetTypeName
            |${errStrings.mkString("\n")}
+           |
```
krzemin, line 84: perhaps additional output line when there's no hint could be avoided?

chimney/src/main/scala/io/scalaland/chimney/internal/macros/MappingMacros.scala
ldrygala: failed on scala 2.12 and 2.11, even with `import scala.collection.compat._`

chimney/src/main/scala/io/scalaland/chimney/internal/DerivationError.scala
ldrygala: 👌

PR #165, opened 2020-04-28
Add support for Scala.js 1.0

liff:
Scala 2.11 with .js 1.0 is excluded because cats and utest are not available for that combination.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,12 @@ jobs:
         jvm:
           - openjdk@8
           - openjdk@1.11
+        scalajs:
+          - 0.6.32
+          - 1.0.1
+        exclude:
+          - scala: 2.11.12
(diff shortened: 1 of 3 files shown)


liff:
Oops, it seems that I got the CI build matrix wrong.

I think it should be:

```yaml
        exclude:
          - scala: 2.11.12
            scalajs: 1.0.1
```

that is, without the list item marker, `-` before `scalajs`.

krzemin:
Thanks, fixed in https://github.com/scalalandio/chimney/commit/4f72a64c69aa9f612893859fd2ff51038aceacf7

liff:
Would it be possible to publish 0.5.1 for Scala.js 1.0?

krzemin:
Sure, I'll publish release later today.

krzemin:
Released 🎉

Issue #171, opened 2020-07-06
Request for cats Ior support


calvinlfer:
Hello all,

Thank you for creating this amazing and useful library. It would be amazing if Chimney supported the cats Ior datatype similar to how Validated support is done via lifted Transformers

Issue #170, opened 2020-07-06
TransformerFSupport for cats-effect Sync


kubukoz:
I would like to have a transformer with effects (or maybe just some validations, but in the same effect as the rest of the logic in the same trait).

Since there's no support for that now, It would be nice to provide `TransformerFSupport` for any `F[_]: Sync`.
Normally `ApplicativeError` would've been enough, but since there's direct usage of iterator methods, they should be suspended in the effect. Unless the API of `TransformerFSupport` is changed, the iterator calls should wrapped in `Sync[F].delay`, and the while loop should be replaced with an appropriate `whileM` or something similar, effect-aware (in comparison to the implementation for Either).

krzemin:
My concern about integration with `Sync` is that it doesn't belong to `cats-core`, but to `cats-effect` which means either depending on it in `chimney-cats` (bad) or introducing another integration module `chimney-cats-effects` (also bad?).

As far as I understand `ApplicativeError`, it allows you to raise/handle single error, but in case of nested wrapped transformers, you may want to be interested in gathering all the errors, so I think AE leads to wrong abstraction here.

There is also a lot of confusion about implementing `traverse` for `TransformerFSupport`. Iterator provided to this method has the underlying data structure already materialized, so there are essentially no effects other than consuming the iterator, which you have guarantee that is local to the `traverse` function. More attention should be paid to provide efficient implementation that doesn't produce any ephemeral garbage.

Either way, PRs welcome 👍

krzemin:
Closing, as lifted transformers are deprecated.

PR #172, opened 2020-07-07
Support Ior

calvinlfer:
Implement cats.Ior support for TransformerF in the cats integration module

Closes https://github.com/scalalandio/chimney/issues/171

### DIFF
--- a/chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
+++ b/chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
@@ -2,10 +2,8 @@ package io.scalaland.chimney
 
 import _root_.cats.data._
 import _root_.cats.kernel.Semigroup
-import _root_.cats.{Apply, Eval}
 
 import scala.collection.compat._
-import scala.collection.compat.immutable.{LazyList => CompatLazyList}
 
 package object cats extends CatsImplicits
 
@@ -41,12 +39,32 @@ trait LowPriorityImplicits {
       override def map[A, B](fa: Ior[EE, A], f: A => B): Ior[EE, B] =
         fa.map(f)
 
-      override def traverse[M, A, B](it: Iterator[A], f: A => Ior[EE, B])(implicit fac: Factory[B, M]): Ior[EE, M] =
-        it.foldRight(Eval.always(pure(CompatLazyList.empty[B]))) { (next: A, acc: Eval[Ior[EE, CompatLazyList[B]]]) =>
-            Apply[Ior[EE, *]].map2Eval(f(next), acc)(_ #:: _)
+      override def traverse[M, A, B](it: Iterator[A], f: A => Ior[EE, B])(implicit fac: Factory[B, M]): Ior[EE, M] = {
+        var leftSide: EE = null.asInstanceOf[EE]
+        val rightSide = fac.newBuilder
+        var isLeft = false
+
+        while (it.hasNext && !isLeft) {
+          f(it.next()) match {
+            case Ior.Left(a) =>
+              if (leftSide == null) leftSide = a
+              else leftSide = Semigroup[EE].combine(leftSide, a)
+              isLeft = true
+
+            case Ior.Right(b) =>
+              rightSide += b
+
+            case Ior.Both(a, b) =>
+              if (leftSide == null) leftSide = a
+              else leftSide = Semigroup[EE].combine(leftSide, a)
+              rightSide += b
           }
-          .value
-          .map(fac.fromSpecific(_))
+        }
+
+        if (isLeft) Ior.left(leftSide)
+        else if (leftSide != null) Ior.both(leftSide, rightSide.result())
+        else Ior.right(rightSide.result())
+      }
     }
 
   implicit def TransformerFValidatedSupport[EE: Semigroup]: TransformerFSupport[Validated[EE, +*]] =
--- a/chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
+++ b/chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
@@ -2,6 +2,8 @@ package io.scalaland.chimney
 
 import _root_.cats.data._
 import _root_.cats.kernel.Semigroup
+import _root_.cats.Traverse
+import _root_.cats.instances.list._
 
 import scala.collection.compat._
 
@@ -15,9 +17,33 @@ trait CatsImplicits extends LowPriorityImplicits {
 
   implicit def TransformerFValidatedNelSupport[E]: TransformerFSupport[ValidatedNel[E, +*]] =
     TransformerFValidatedSupport[NonEmptyList[E]](implicitly)
+
+  implicit def TransformerFIorNecSupport[E]: TransformerFSupport[IorNec[E, +*]] =
+    TransformerFIorSupport[NonEmptyChain[E]](implicitly)
+
+  implicit def TransformerFIorNelSupport[E]: TransformerFSupport[IorNel[E, +*]] =
+    TransformerFIorSupport[NonEmptyList[E]](implicitly)
+
+  implicit def TransformerFIorNesSupport[E]: TransformerFSupport[IorNes[E, +*]] =
+    TransformerFIorSupport[NonEmptySet[E]](implicitly)
 }
 
 trait LowPriorityImplicits {
+
+  implicit def TransformerFIorSupport[EE: Semigroup]: TransformerFSupport[Ior[EE, +*]] =
+    new TransformerFSupport[Ior[EE, +*]] {
+      override def pure[A](value: A): Ior[EE, A] =
+        Ior.right(value)
+
+      override def product[A, B](fa: Ior[EE, A], fb: => Ior[EE, B]): Ior[EE, (A, B)] =
+        fa.flatMap(a => fb.map(b => (a, b)))
+
+      override def map[A, B](fa: Ior[EE, A], f: A => B): Ior[EE, B] =
+        fa.map(f)
+
+      override def traverse[M, A, B](it: Iterator[A], f: A => Ior[EE, B])(implicit fac: Factory[B, M]): Ior[EE, M] =
+        Traverse[List].traverse[Ior[EE, *], A, B](it.toList)(f).map(fac.fromSpecific(_))
+    }
 
   implicit def TransformerFValidatedSupport[EE: Semigroup]: TransformerFSupport[Validated[EE, +*]] =
     new TransformerFSupport[Validated[EE, +*]] {
--- a/chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
+++ b/chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
@@ -2,10 +2,10 @@ package io.scalaland.chimney
 
 import _root_.cats.data._
 import _root_.cats.kernel.Semigroup
-import _root_.cats.Traverse
-import _root_.cats.instances.list._
+import _root_.cats.{Apply, Eval}
 
 import scala.collection.compat._
+import scala.collection.compat.immutable.{LazyList => CompatLazyList}
 
 package object cats extends CatsImplicits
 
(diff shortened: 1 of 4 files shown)


chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
```diff
@@ -15,10 +17,34 @@ trait CatsImplicits extends LowPriorityImplicits {
...
+        Ior.right(value)
+
+      override def product[A, B](fa: Ior[EE, A], fb: => Ior[EE, B]): Ior[EE, (A, B)] =
+        fa.flatMap(a => fb.map(b => (a, b)))
+
+      override def map[A, B](fa: Ior[EE, A], f: A => B): Ior[EE, B] =
+        fa.map(f)
+
+      override def traverse[M, A, B](it: Iterator[A], f: A => Ior[EE, B])(implicit fac: Factory[B, M]): Ior[EE, M] =
+        Traverse[List].traverse[Ior[EE, *], A, B](it.toList)(f).map(fac.fromSpecific(_))
```
kubukoz, line 45: There might be some laziness concerns in converting the iterator to a list. Maybe we can go through a lazy list?

If that works, then we can probably generalize the entire instance to Applicative, similarly to what I wanted to do in #170

calvinlfer: Thanks to some direction from @kubukoz, we can do this lazily using

```scala
override def traverse[M, A, B](it: Iterator[A], f: A => Ior[EE, B])(implicit fac: Factory[B, M]): Ior[EE, M] =
  it.foldRight(Eval.always(pure(CompatLazyList.empty[B]))) { (next: A, acc: Eval[Ior[EE, CompatLazyList[B]]]) =>
      Apply[Ior[EE, *]].map2Eval(f(next), acc)(_ #:: _)
  }.value.map(fac.fromSpecific(_))
```

krzemin: Is `it.foldRight` intentional in your implementation? In my understanding it destroys any laziness concerns.

chimneyCats/src/test/scala/io/scalaland/chimney/cats/CatsValidatedSpec.scala
```diff
@@ -16,7 +16,7 @@ object CatsValidatedSpec extends TestSuite {
 
   val tests = Tests {
 
-    "transform always succeeds" - {
+    test("transform always succeeds") - {
```
krzemin, line 19: Any reason for using different syntax than the rest of the library?

krzemin:
@calvinlfer thanks for your PR! It would be nice if you could add subpage in docs about `Ior` integration, similar to [what we currently have for `Validated`](https://scalalandio.github.io/chimney/transformers/cats-integration.html#validated-support-for-lifted-transformers)

chimneyCats/src/test/scala/io/scalaland/chimney/cats/CatsValidatedSpec.scala
calvinlfer: Ah sorry, I should have converted the rest of them as well, I was getting deprecated syntax for the existing tests so I wanted to move to the new syntax. Hope that’s alright

calvinlfer:
Sounds great @krzemi!, @kubukoz and I were trying to derive a TransformerF for any effect that had an Applicative instance but I think we can come around in another PR for that

chimneyCats/src/test/scala/io/scalaland/chimney/cats/CatsValidatedSpec.scala
krzemin: No worries, the best would be to switch to the new syntax in a separate PR.

chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
calvinlfer: So I've mimicked the [Cats implementation]((https://github.com/typelevel/cats/blob/master/core/src/main/scala-2.13%2B/cats/instances/lazyList.scala#L61)) for traverse on lazy list. 
As far as I understand, Cats is using Eval to be lazy but I could be wrong

calvinlfer: Initially, I used a tail-recursive function (for early termination) with a Vector accumulator (as I needed to append) and terminated as soon as I had a Ior.Left and then at the very, I can switched it to the builder. Would you like me to switch back to this?

chimneyCats/src/test/scala/io/scalaland/chimney/cats/CatsValidatedSpec.scala
calvinlfer: Sounds good! I'll switch to the new syntax in a separate PR

chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
calvinlfer: @krzemin 
```scala
      override def traverse[M, A, B](it: Iterator[A], f: A => Ior[EE, B])(implicit fac: Factory[B, M]): Ior[EE, M] = {
        @tailrec
        def go(rem: Iterator[A], acc: Ior[EE, Vector[B]]): Ior[EE, M] =
          rem.nextOption() match {
            case Some(a) =>
              val newAcc = Apply[Ior[EE, *]].map2(acc, f(a))(_ :+ _)
              go(rem, newAcc)

            case None =>
              println(acc)
              acc.map(fac.fromSpecific(_))
          }
        go(it, Ior.right(Vector.empty))
      }
```

calvinlfer: That mimicks the same semantics as foldRight + Eval, let me know if you want me to swap it out for this

krzemin: Ok, let's clarify few things:
- laziness in case of accessing provided `Iterator` is not very much a concern, as in most of the cases it contains data already present in memory (case classes fields)
- in fact the problem was that intermediate data structure (a `List`) was created from the iterator, leading to unnecessary garbage
- your tailrec implementation is very much similar to `foldLeft`, but the intermediate `Vector` is used for buffering (appending) results, that leads to unnecessary garbage as well; `LazyList` based implementation has the same issue 

Intermediate garbage can be avoided by:
1. consuming the iterator elements directly
2. using scala collections builder directly

Examples of such `traverse` implementations are available [for `Either`](https://github.com/scalalandio/chimney/blob/master/chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala#L134) and [for `Option`](https://github.com/scalalandio/chimney/blob/master/chimney/src/main/scala/io/scalaland/chimney/TransformerFSupport.scala#L88).

calvinlfer: Ah I understand what you are saying now @krzemin and I realized the tail-recursive implementation didn't terminate early, Here's the improved version:
```scala
      override def traverse[M, A, B](it: Iterator[A], f: A => Ior[EE, B])(implicit fac: Factory[B, M]): Ior[EE, M] = {
        @tailrec
        def go(rem: Iterator[A], acc: Ior[EE, Vector[B]]): Ior[EE, M] =
          rem.nextOption() match {
            case Some(a) =>
              val newAcc = Apply[Ior[EE, *]].map2(acc, f(a))(_ :+ _)
              
              if (newAcc.isLeft) newAcc.map(fac.fromSpecific(_))
              else go(rem, newAcc)

            case None =>
              acc.map(fac.fromSpecific(_))
          }
        go(it, Ior.right(Vector.empty))
      }
```

Let me take a look at the traverse for Either and Option that you provided and see if I can get rid of the intermediate Vector

calvinlfer: Hello @krzemin, sorry to keep bothering you, I took a look at the implementations for Either and Option. I noticed in the `Either` implementation, you have a type constraint on the error result `C[X] <: IterableOnce[X]`. However, I don't see a way to get rid of the intermediate `Vector` in the general case for `Ior` specifically since this allows us to support any `EE` that has a `Semigroup`instance as opposed to limiting the left side to `IterableOnce`.

calvinlfer:
> @calvinlfer thanks for your PR! It would be nice if you could add subpage in docs about `Ior` integration, similar to [what we currently have for `Validated`](https://scalalandio.github.io/chimney/transformers/cats-integration.html#validated-support-for-lifted-transformers)

I have added documentation now :scroll:

chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
krzemin: Why can't you just have `var ee: EE` and accumulate on it with Semigroup? In the while loop, depending on what you get, you can:
- Left -> terminate early with the EE value got here
- Both -> accumulate warning on `ee` and add the element to the builder
- Right -> just add the element to the builder

After the loop, if you didn't terminate early, you need to check if you need to return Both or Right (in both cases you call `.result()` on builder to obtain final collection).

I totally agree that we shouldn't limit to `IterableOnce` as it's not the cats' way.

calvinlfer: Oops @krzemin turns out that I have forgotten how to write mutable code :laughing:, this works:
```scala
      override def traverse[M, A, B](it: Iterator[A], f: A => Ior[EE, B])(implicit fac: Factory[B, M]): Ior[EE, M] = {
        var leftSide: EE = null.asInstanceOf[EE]
        val rightSide = fac.newBuilder
        var earlyTermination = false

        while (it.hasNext && !earlyTermination) {
          f(it.next()) match {
            case Ior.Left(a) =>
              if (leftSide == null) leftSide = a
              else leftSide = Semigroup[EE].combine(leftSide, a)
              earlyTermination = true

            case Ior.Right(b) =>
              rightSide += b

            case Ior.Both(a, b) =>
              if (leftSide == null) leftSide = a
              else leftSide = Semigroup[EE].combine(leftSide, a)
              rightSide += b
          }
        }

        if (earlyTermination) Ior.left(leftSide)
        else if (leftSide != null) Ior.both(leftSide, rightSide.result())
        else Ior.right(rightSide.result())
      }
```

krzemin: Hahaha, exactly what I meant 😄

calvinlfer: hehe sweet! thank you so much @krzemin 
I'll incorporate this in

calvinlfer:
All good to go @krzemin :rocket:

chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
```diff
@@ -15,10 +18,58 @@ trait CatsImplicits extends LowPriorityImplicits {
...
+      override def traverse[M, A, B](it: Iterator[A], f: A => Ior[EE, B])(implicit fac: Factory[B, M]): Ior[EE, M] = {
+        var leftSide: EE = null.asInstanceOf[EE]
+        val rightSide = fac.newBuilder
+        var isLeft = false
+
+        while (it.hasNext && !isLeft) {
+          f(it.next()) match {
+            case Ior.Left(a) =>
+              if (leftSide == null) leftSide = a
+              else leftSide = Semigroup[EE].combine(leftSide, a)
```
krzemin, line 54: Assuming early termination, you don't need to check if `leftSide == null` - it will always be true.

chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
```diff
@@ -2,8 +2,11 @@ package io.scalaland.chimney
 
 import _root_.cats.data._
 import _root_.cats.kernel.Semigroup
+import _root_.cats.{Apply, Eval}
 
+import scala.annotation.tailrec
 import scala.collection.compat._
+import scala.collection.compat.immutable.{LazyList => CompatLazyList}
```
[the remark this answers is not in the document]
krzemin: applies to other imports (Apply/Eval/tailrec) too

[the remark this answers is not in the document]
calvinlfer: Oops, will remove

chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
calvinlfer: So that’s true if you have Rights followed by a Left but this is not the case for Boths followed by a Left because the left side will have accumulated data and we would need to combine the data from the first Left before stopping

krzemin: Oh yes, you're right - assuming you want to return also accumulated data, not only the last Left's one.

chimneyCats/src/main/scala/io/scalaland/chimney/cats/package.scala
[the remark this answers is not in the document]
calvinlfer: Fixed my good sir!

calvinlfer:
Thank you so much for all your help and patience @krzemin and @kubukoz  :blue_heart:

krzemin:
Thank you @calvinlfer for your contribution!

calvinlfer:
Thank you @krzemin! 😀

Issue #173, opened 2020-07-07
Possible bug: can't use `withCoproductInstance` on a `TransformerF.define`


kubukoz:
Again, thanks for the great library!

I already asked a question on the gitter, but I'll repost it here as it didn't get much attention.

I tried defining a `TransformerF` for the first time:

```scala
sealed trait Foo
case object Bar extends Foo
case object Baz extends Foo

sealed trait Foo2
case object Bar2 extends Foo2
case object Baz2 extends Foo2

val transformer = TransformerF
  .define[Option, Foo, Foo2]
  .withCoproductInstanceF((_: Bar.type) => Bar2.some)
  .withCoproductInstanceF((_: Baz.type) => Baz2.some)
  .buildTransformer
```

this compiles fine, but...
if I replace even one withCoproductInstanceF with a non-effectful one:

```scala
val transformer = TransformerF
  .define[Option, Foo, Foo2]
  .withCoproductInstance((_: Bar.type) => Bar2)
  .withCoproductInstanceF((_: Baz.type) => Baz2.some)
  .buildTransformer
```

(btw. the `.some` part is cats syntax)

I get a compilation error:

```
[E]      type mismatch;
[E]        Option[Foo2]|Foo2
[E]      L149:     .buildTransformer
[E]                 ^
```

It looks to me like the error appears while expanding the macro (or in the code it's expanded to, in fact).

I would expect that example to work though, as from the user side everything typechecks (until the `buildTransformer` call).

Is it a problem with Chimney, or am I doing something wrong?

Chimney version: 0.5.2

krzemin:
Hi @kubukoz, thanks for reporting. Your usage of the library seems to be correct, it really looks like bug covering some strange corner case. I will look deeper into this soon.

krzemin:
Resolved with #174

PR #174, opened 2020-07-08
fix issue #173

krzemin:
As reported in #173, library generated incorrect code when transforming coproducts (`withCoproductInstance`) mixed with wrapped transformers. The fix was to omit explicit cast when accessing non-wrapped function provided by the user. Original example from #173 is included as a new test.


krzemin:
@ldrygala In current implementation the order doesn't matter as underlying data structure in the macro will be the same. But added suggested test for the sake of regression.

PR #178, opened 2020-08-11
Prefer TransformerF derivation in lifted transformations

### LINKED ISSUE #177: Implicit TransformerF does not override derivation in nested transformation
It appears that the macro derivation does not attempt to locate a `TransformerF` instance for nested transformations and will only perform derivation as well when the source and destination types have matching structures.

I noticed this when trying too convert a case class to the `sealed abstract case class` trick to control initialization which then prevented my existing `transformIntoF` from instantiating it directly. Actually, the fact that the transformer was skipped is what prompted me to do that in the first place to prevent missing implicits from generating working code.

The following example fails to compile with `class Bar is abstract; cannot be instantiated`, at least on Scala 2.13.3 with Chimney 0.5.3:

```scala
package test

import io.scalaland.chimney.TransformerF

object foobar {

  case class Foo(x: Int, y: String)

  sealed abstract case class Bar(x: Int, y: String)
  object Bar {
    def make(x: Int, y: String): Bar = new Bar(x, y) {}
  }

  implicit val fooToBar: TransformerF[Option, Foo, Bar] =
    f => Some(Bar.make(f.x, f.y))

  // error: class Bar is abstract; cannot be instantiated
  (1, List(Foo(1, "test"))).transformIntoF[Option, (Int, List[Bar])]

  // these all work correctly however
  Foo(1, "test").transformIntoF[Option, Bar]
  List(Foo(1, "test")).transformIntoF[Option, List[Bar]]
  (1, Foo(1, "test")).transformIntoF[Option, (Int, Bar)]

}
```

The case can also be tested without the `sealed abstract` trick, but becomes more subtle and a potential source of bugs if untested:

```scala
scala> case class Foo(x: Int, y: String)
scala> case class Bar(x: Int, y: String)
scala> implicit val t: TransformerF[Option, Foo, Bar] = f => Some(Bar(f.y.length, f.x.toString)) // Swapped
scala> (1, List(Foo(1, "test"))).transformIntoF[Option, (Int, List[Bar])]
val res1: Option[(Int, List[Bar])] = Some((1,List(Bar(1,test)))) // Wrong! expected Some((1,List(Bar(4,"1"))))
```

willtrnr:
Resolves #177 

Changes lifted transformation to recursively derive a `TransformerF` instances.

This fixes the case where a type that can be derived as a `Transformer` has a member which is also derivable as a `Transformer`, but has an implicit `TransformerF` instance available. In such a case the containing type will short circuit to the non-lifted variant and prevent that sub-tree from using its implicit lifted instance since derivation will no longer use the wrapper type.

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
@@ -644,15 +644,8 @@ trait TransformerMacros extends TransformerConfiguration with MappingMacros with
         case (None, Some(localImplicitTree)) =>
           Right(TransformerBodyTree(localImplicitTree.callTransform(srcPrefixTree), isWrapped = false))
         case (None, None) =>
-          def deriveTransformer =
-            deriveTransformerTree(srcPrefixTree, recConfigNoWrapper)(From, To)
-              .mapRight(tree => TransformerBodyTree(tree, isWrapped = false))
-
-          def deriveTransformerF =
-            deriveTransformerTree(srcPrefixTree, recConfig)(From, To)
-              .mapRight(tree => TransformerBodyTree(tree, isWrapped = true))
-
-          deriveTransformer rightOrElse deriveTransformerF
+          deriveTransformerTree(srcPrefixTree, recConfig)(From, To)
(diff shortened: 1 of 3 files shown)


krzemin (approved):
LGTM, thank you 👍

PR #179, opened 2020-08-12
WIP Add support of injection custom derivation rules

ulanzetz:
Sometimes user needs to add custom transformer rules, which cannot be defined by implicit level. For example, Option[List[A]] -> List[B], or customize library order of transformation rules. This solution will allow users to customize it by adding their own rules. Users, who use their own forks with changed rules, will be able migrate to upstream and add their own rules into their projects. 

WDYT, @krzemin?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -75,8 +75,8 @@ lazy val root = project
   .settings(settings: _*)
   .settings(publishSettings: _*)
   .settings(noPublishSettings: _*)
-  .aggregate(chimneyJVM, chimneyJS, chimneyCatsJVM, chimneyCatsJS)
-  .dependsOn(chimneyJVM, chimneyJS, chimneyCatsJVM, chimneyCatsJS)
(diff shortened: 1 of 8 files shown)


krzemin:
That's interesting proposal, but it push users to use low level macro api in order to define their own rules.

I do think the proper solution would be to fix the implicits.

ulanzetz:
If derivation rule depends on transformer derived by macro, it cannot be defined by implicits. (or use something like DependentTransformer defined in #167, but it will not work, if transformer depends on several transformers).

krzemin:
> If derivation rule depends on transformer derived by macro, it cannot be defined by implicits.

This is true only for current implementation, because we disable macro expansion when inferring implicit. I would like to have it enabled, but then we would need to carefully track for recursive macro invocations for the same types. Definitely something not trivial, but I would like to investigate what could be done here.

Issue #166, opened 2020-05-04
Ability to not use auto-derivation


pjrt:
Auto derivation, though convenient, can cause increases in compile time (since the macro runs on a per-call-site basis), as well as make implicit instances of Transformers placed in the companion object of a type useless (causes an implicit resolution conflict). 

A lot of libraries (like Pureconfig and Scanamo) have moved away from auto derivation because of it, and instead allowing the user to call a `semi` and `auto` package, which they can use as they please (semi usually used in the companion of a type). 

It would be great if Chimney followed this pattern as well.

mrobakowski:
I don't know if I should open an additional issue for that, but since I believe this issue is very related. `Transformer`s in companion objects _sometimes_ work in _some_ version of chimney, which is very annoying. For example this works in 0.5.3, but doesn't work in 0.6.0.

```scala
case class Id(value: String)
object Id {
  implicit val idToLong: TransformerF[Option[+*], Id, Long] = id => Either.catchOnly[NumberFormatException](id.value.toLong).toOption
}

case class Foo(id: Id)
case class Bar(id: Long)

val foo = Foo(Id("1"))
val bar = foo.transformIntoF[Option[+*], Bar]
```
https://scastie.scala-lang.org/sMQz9zqzTemTMwJgeJNwWw

Curiously, if we try to `transformF` an `Id` into `Long`, it doesn't work in both versions. It's a bit of a mess :joy:

Dr-Nikson:
Any updates on it? 
With auto-derivation it's almost impossible to catch bugs at compile time, it feels like implicit type conversions :(

MateuszKubuszok:
I am willing to fix it, but none of the obvious solutions I find to be satisfying.

The problem is that Chimney attempts optimizations no other library (besides Jsoniter) attempts to do and that it allows customizations that no other derivation-based library has. What do I mean by that?

Let's say we have:

```scala
case class InnerFrom(a: Int)
case class OuterFrom(inner: InnerFrom)

case class InnerTo(a: String)
case class OuterTo(inner: InnerTo)

implicit val intToString: Transformer[Int, String] = _.toString
```

Let's say Chimney behaves the same way as any other way as any other library, e.g. Circe, when it comes to automatic derivation. Then when we'd do:

```scala
OuterFrom(InnerFrom(10)).transformInto[OuterTo]
```

compiler would generate something like:

```scala
OuterFrom(InnerFrom(10)).transformInto[OuterTo]({
[...]

PR #181, opened 2020-08-27
Make Transformer auto-derivation pluggable via import

Leammas:
This is breaking, but IMO required, change providing fine-grained control over `Transformer` derivation.
The problem stated in the issue https://github.com/scalalandio/chimney/issues/166

What do you think of it? Should I make a separate section in the docs if it's applicable?

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ rules and transform your objects with as little boilerplate as possible.
 
 ```scala
 import io.scalaland.chimney.dsl._
+import io.scalaland.chimney.derivation.auto._
 
 val event = command.into[CoffeeMade]
   .withFieldComputed(_.at, _ => ZonedDateTime.now)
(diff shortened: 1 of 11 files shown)


krzemin:
I agree, this change is highly needed. However we can make it a bit less breaking for the users by:

- keeping `derive` method in `Transformer` as it is (just without implicit); it won't break current usages and there won't be need for `TransformerDerivation` trait
- keeping current `dsl._` package working with autoderivation; introduce `dsl.semiauto._` for DSL without these implicits (perhaps extracting some `DslBase` trait will be useful)
- this way it should be backwards compatible and allow to opt-out from autoderivation; to verify this, I recommend reverting any changes to the current test set and introduce new test suite where `dsl.semiauto` is used to:
  - verify that usages relying on auto don't compile
  - verify that implicit resolution conflicts under `dsl._` compile under `dsl.semiauto._`
- please apply the same logic to `TransformerF` and `Patcher`

mrobakowski:
Any updates on this? I found a regression in `0.6.0` that _maybe_ could be resolved by this PR. https://github.com/scalalandio/chimney/issues/166#issuecomment-736567028

Leammas:
I'll try harder to work on it this week. Feel free to catch over if I'm unable to invest the time

Leammas:
Sorry, I believe I will not be able to finish it in a reasonable time.

PR #183, opened 2020-10-01
Enable support of macro-dependent implicit transformers

### LINKED ISSUE #158: Unable to compile when transforming Option[Seq[T] to Seq[T]

ulanzetz:
Use inferImplicitValue with option withMacrosDisabled = false. To avoid recursive deriving it's check that strings "Transformer.derive" and "TransformerF.derive" is not in found tree. Maybe there is more elegant way to check usage of this methods, but it works too.

It will resolve #158 and similar issues, when we try to use implicit transformer dependent of other, witch should be derived by macro.

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
@@ -687,13 +687,26 @@ trait TransformerMacros extends TransformerConfiguration with MappingMacros with
       silent = true,
       mode = c.TYPEmode,
       withImplicitViewsDisabled = true,
-      withMacrosDisabled = true
+      withMacrosDisabled = false
     )
 
     scala.util
-      .Try(c.inferImplicitValue(tpeTree.tpe, silent = true, withMacrosDisabled = true))
+      .Try(c.inferImplicitValue(tpeTree.tpe, silent = true, withMacrosDisabled = false))
       .toOption
-      .filterNot(_ == EmptyTree)
+      .filterNot { tree =>
+        tree == EmptyTree ||
+        isDeriving(tree)
+      }
+  }
+
+  private def isDeriving(tree: Tree): Boolean = {
+    tree match {
+      case TypeApply(Select(qualifier, name), _) =>
+        qualifier.tpe =:= weakTypeOf[io.scalaland.chimney.Transformer.type] && name.toString == "derive"
+      case Apply(TypeApply(Select(qualifier, name), _), _) =>
+        qualifier.tpe =:= weakTypeOf[io.scalaland.chimney.TransformerF.type] && name.toString == "derive"
+      case _ => false
+    }
   }
 
   private def notSupportedDerivation(
--- a/chimney/src/test/scala/io/scalaland/chimney/DslSpec.scala
+++ b/chimney/src/test/scala/io/scalaland/chimney/DslSpec.scala
@@ -912,6 +912,38 @@ object DslSpec extends TestSuite {
         Bar1(1, Baz(Some(Bar1(2, Baz(None))))).transformInto[Bar2] ==> Bar2(Baz(Some(Bar2(Baz(None)))))
       }
     }
+
+    "support macro dependent transformers" - {
+      "Option[List[A]] -> List[B]" - {
+        implicit def optListT[A, B](implicit underlying: Transformer[A, B]): Transformer[Option[List[A]], List[B]] =
+          _.toList.flatten.map(underlying.transform)
+
+        case class ClassA(a: Option[List[ClassAA]])
+        case class ClassB(a: List[ClassBB])
+        case class ClassC(a: List[ClassBB], other: String)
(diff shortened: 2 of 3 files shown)


chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
```diff
@@ -687,13 +687,17 @@ trait TransformerMacros extends TransformerConfiguration with MappingMacros with
...
     )
 
     scala.util
-      .Try(c.inferImplicitValue(tpeTree.tpe, silent = true, withMacrosDisabled = true))
+      .Try(c.inferImplicitValue(tpeTree.tpe, silent = true, withMacrosDisabled = false))
       .toOption
-      .filterNot(_ == EmptyTree)
+      .filterNot { tree =>
+        tree == EmptyTree ||
+        tree.toString.contains("Transformer.derive") ||
```
krzemin, line 698: You have just blown my mind 💥 I'm glad that we have such a simple workaround that seems to cover all the current test suites 👍 

I'm just afraid that when user defines:
```scala
object Transformer {
  implicit val derive: Transformer[...] = ...
}
```
it won't be selected by the library, even if there are no macros involved. That's just only one of many consequences of relying on string filtering.

More reliable method of preventing generation of code that self-references, would be to keep a compile-time stack in `ThreadLocal`, which contains type pairs for which we run our own macro. At the macro invocation we can just lookup the stack and refuse to generate an instance, when there is a type pair on the stack. Quite similar to what [magnolia does](https://github.com/propensive/magnolia/blob/e10e0b153501664987ed07316d150070555e6b4d/core/shared/src/main/scala/magnolia.scala#L801).

krzemin (commented):
Let me just collect all the issues that relies on this problem that way or another. Based on them, we can add more test cases to see if your solution covers all of them. I'm OK with merging the string-based solution if we don't find any serious problems with it.

Alternatively, we can still explore ThreadLocal+stack based solution, as it tends to be more reliable.

chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
ulanzetz: @krzemin, I found solution without string-based checking. Exactly targeted to to derive methods of io.scalaland.chimney.Transformer and io.scalaland.chimney.TransformerF companions

krzemin: Good 👍 That should be sufficiently reliable IMO.

On the other side, we still perform implicit macro expansion when looking for implicit, so it may cause some performance regression. But it may be fixed later, if needed.

chimney/src/test/scala/io/scalaland/chimney/DslSpec.scala
```diff
@@ -912,6 +912,38 @@ object DslSpec extends TestSuite {
         Bar1(1, Baz(Some(Bar1(2, Baz(None))))).transformInto[Bar2] ==> Bar2(Baz(Some(Bar2(Baz(None)))))
       }
     }
+
+    "support macro dependent transformers" - {
```
krzemin, line 916: Similar test for `TransformerF` would be much appreciated.

ulanzetz: Added test for TranformerF with safe option unwrapping rule. Declined #155, cause now such rule can be defined by implicits

chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
ulanzetz: Maybe add flag withMacrosDisabled in config with default false value? Users who will not use this feature will be able to disable it and safe compile performance

krzemin (approved):
Looks good 👍 Thank you!

Issue #182, opened 2020-09-14
Transformation fail on nestes sealed trait structure


matwojcik:
This works:
```scala

object foo {

  sealed trait A extends Product with Serializable
  case object A1 extends A

  object into {
    sealed trait A extends Product with Serializable
    case object A1 extends A
  }

  def convert(a: A): into.A = a.transformInto[into.A]
}
```

But this fails:
```scala
object foo {

  sealed trait A extends Product with Serializable
  sealed trait AA extends A
  case object A1 extends AA

  object into {
    sealed trait A extends Product with Serializable
    sealed trait AA extends A
    case object A1 extends AA
  }

  def convert(a: A): into.A = a.transformInto[into.A]
}
```

```
Chimney can't derive transformation from foo.A to foo.into.A
[error] 
[error] foo.into.A
[error]   can't transform coproduct instance foo.AA to foo.into.A
[error] 
[error] 
[error] Consult https://scalalandio.github.io/chimney for usage examples.
[error] 
[error]   def convert(a: A): into.A = a.transformInto[into.A]
```

krzemin:
Resolved by [@ulanzetz's PR](#187).

PR #187, opened 2020-10-06
Fix issue #182

ulanzetz:


krzemin (commented):
Good job, thank you 👍

Would that enable transformations between flat and deep sealed families? If yes, I would add it to the test suite.

Maybe we can update the [docs page](https://scalalandio.github.io/chimney/transformers/customizing-transformers.html#transforming-coproducts) and note explicitly that deep sealed hierarchies are supported?

ldrygala (approved):
Great job 🚀

ulanzetz:
> Good job, thank you 
> 
> Would that enable transformations between flat and deep sealed families? If yes, I would add it to the test suite.
> 
> Maybe we can update the [docs page](https://scalalandio.github.io/chimney/transformers/customizing-transformers.html#transforming-coproducts) and note explicitly that deep sealed hierarchies are supported?

Added tests and updated docs, but I think it will be unobvious if something like this will be available
```
object From {
  sealed trait Base

  sealed trait BranchA extends Base
  sealed trait BranchB extends Base

  case object FirstLeaf extends BranchA
  case object SecondLeaf extends BranchB
}

object To {
  sealed trait Base

  sealed trait BranchA extends Base
  sealed trait BranchB extends Base

  case object FirstLeaf extends BranchB
  case object SecondLeaf extends BranchA
}

(From.FirstLeaf: From.Base).transformInfo[To.Base] ==> To.FirstLeaf
```
Maybe allow only transformations, when list of destination parents is prefix of list of source parents and order of parents is linear?

docs/source/transformers/customizing-transformers.rst
```diff
@@ -173,3 +173,6 @@ do that. Let's convert any ``Channel.Alpha`` to ``Color.Blue``.
 
 After providing a default, Chimney can prove the transformation
 is total and use provided function, when it's needed.
+
+Transformations between flat sealed trait hierarchies and deep trait
+hierarchies containing nested sealed traits is also available.
```
krzemin, line 178: is -> are

chimney/src/test/scala/io/scalaland/chimney/IssuesSpec.scala
```diff
@@ -491,3 +495,19 @@ object tag {
     def apply[T](t: T): T @@ U = t.asInstanceOf[T @@ U]
   }
 }
+
+object foo {
```
krzemin, line 499: maybe we can have it in `examples` package?

krzemin:
> Maybe allow only transformations, when list of destination parents is prefix of list of source parents and order of parents is linear?

Hmm, I'm not sure if your example causes any issues on deriving a transformer. Well, we can have different inheritance paths, but as long as we match by name and able to construct transformer, it should be fine?

---

I'd be worried about slightly different case:

```scala
object From {
  sealed trait Base
  case object Leaf extends Base
}

object To {
  sealed trait Base
  case object Leaf extends Base
  object Inner {
    case object Leaf extends Base
  }
}
```
Which target leaf would be returned as a result? `To.Leaf` or `To.Inner.Leaf`? I think that currently we don't resolve such conflicts. But that's a separate issue, not really related to this PR. Suggest creating another issue+PR and failing early with derivation error in case we find conflicting names in the target sealed hierarchy.

ulanzetz:
>To.Leaf or To.Inner.Leaf
Nondeterministic. Maybe fail, if more than one subclasses with same name?

krzemin:
> Nondeterministic. Maybe fail, if more than one subclasses with same name?

If they are defined with the same name in the same scope, the compiler just fails with an error. Otherwise (as in the example), it depends on in which order `knownDirectSubclasses` call returns them. Other than that, we are currently deterministic I think. But we can improve on that.

krzemin (approved):
LGTM 👍

ulanzetz:
@krzemin knownDirectSubclasses returns Set of Symbol, if there are less then 5 different subclasses, list of toInstances will bу ordered and choice will be deterministic. But other it will be different for each compilation
```
object Test {
  object From {
    sealed trait Base

    case object Leaf extends Base
  }

  object To {
    sealed trait Base

    case object Leaf extends Base

    object a {
      case object Leaf extends Base
    }

    object b {
      case object Leaf extends Base
    }

    object c {
      case object Leaf extends Base
    }

    object d {
      case object Leaf extends Base
    }

    object e {
      case object Leaf extends Base
    }
  }

  (From.Leaf: From.Base).transformInto[To.Base]
}
```
I have compiled two times, first time `To.c.Leaf.type` was chosen, second time - `To.b.Leaf.type`

krzemin:
Created issue for that specific improvement: https://github.com/scalalandio/chimney/issues/188

kubukoz:
@krzemin are you planning a release soon? There's a bunch of stuff pending ;)

krzemin:
Hoping for early/mid this week, together with #190 when it's ready.

kubukoz:
Cool, cool, thanks!

Issue #176, opened 2020-08-05
Control over default configurations of transformers


d12frosted:
Hello,

I have a use case where I want chimney by default not to support default values of case classes and make this feature opt-in, instead of opt-out. I see that in the way chimney currently [generates][1] (derives) transformers it's impossible to do as it simply reuses defaults.

Are there any plans on supporting changing the defaults? 

Thanks in advance.

  [1]: https://github.com/scalalandio/chimney/blob/e9aa7a15df84028ee5df69ab0dd9a7fe2d3fd6f7/chimney/src/main/scala/io/scalaland/chimney/internal/macros/ChimneyBlackboxMacros.scala#L43

krzemin:
It's true that currently there's no possibility to have a global configuration of flags. They are now set in stone and can be changed only one-way using DSL's `enableXXX`/`disableXXX`.

However that should be quite easy to implement such a flexibility. That would require following steps:
- separate configuration flags from field values/providers
- introduce some type that holds configuration and DSL that allows its manipulation; in the companion object there should be a default implicit instance, but user can define own default in the implicit scope
- `.into` and `.derive` will need to take the implicit configuration object and respect its settings
- `Transformer*Into` dsl needs to allow changing flags in both ways (in order to override defaults in the call site, whatever they are)

That would enable opt-in default values and much more.

d12frosted:
Thank you for such a detailed response. I was thinking about similar solution. 

Just for the sake of 'completeness', I think [ConfigDsl][1] should also support changing each flag in both ways.

Now a pragmatic question - is someone willing to implement this in near future? 😸 If no, I could take it.

  [1]: https://github.com/scalalandio/chimney/blob/master/chimney/src/main/scala/io/scalaland/chimney/dsl/ConfigDsl.scala

krzemin:
@d12frosted Feel free to take it 👍

nilskp:
I'd like to add my 2 cents here. I've just spent a couple of days debugging a strange error that ultimately turned out to be a transformation of an `Option` being silently turned the default value of a non-`Option`.

As a general approach, I think it's unfortunate that the default behavior is something other than super strict. I'm now left wondering if I have other bugs waiting to happen.

I think it would be advisable to reconsider the default behavior before 1.0 is released, whenever that'll happen.

nilskp:
[...]

PR #190, opened 2020-10-09
Support for scope-default transformer configuration

krzemin:
This PR addresses #176. It's result of several prototyping rounds made in cooperation with @d12frosted.

More or less it implements the plan sketched here: https://github.com/scalalandio/chimney/issues/176#issuecomment-669329379 

TODO:
- [X] more test cases, especially of newly added flags combinations
- [X] ~~provide analogous mechanism for patchers~~ let's have it as a separate issue (https://github.com/scalalandio/chimney/issues/191)
- [X] rethink naming and placement of newly added types
- [x] update docs


d12frosted (approved):
LGTM, next step would be to pass default flags from top to the very bottom 💯 

Thank you for taking your time to try this idea.

krzemin:
> next step would be to pass default flags from top to the very bottom

@d12frosted it's already supported, see these test cases: https://github.com/scalalandio/chimney/pull/190/files#diff-8bf852c2438394dc69a2bd644c2a815eR1017 Or do you mean something else?

d12frosted:
🤔 yeah, you are right. For some reason I was confused since I didn't see ability to pass configuration to `Transformer.define`. But looking at the tests - seems like it works 😸 thanks

krzemin:
> For some reason I was confused since I didn't see ability to pass configuration to Transformer.define.

Good point, let me elaborate.

Initially I thought that capturing implicit flags on `Transformer.define` and encoding them in the `Transformer(F)Defnition` would be a simple and clean idea. But to capture types passed by implicit instance, we would need third type parameter (for flags) and just calling it as `Transformer.define[From, To]` would not be possible any more.

I could fix it by making `.define` a whitebox macro and look for implicit inside, but... there is a simpler solution: 
- build `Transformer(F)Definition` flags on top of `TransformerFlags.Default`
- require scoped flags only in `.buildTransformer`
- merge both flags stacks after capturing both types

It keeps almost the same semantics, but bring less troubles for the users.

d12frosted:
@krzemin ah I see. This is something I didn't consider in my PRs. Nicely done!

ulanzetz (approved):

d12frosted:
This is wonderful! Thank you @krzemin 🎉

krzemin:
Thank you for your efforts too @d12frosted!

PR #197, opened 2020-11-13
WIP Resolving TransformerF and Transformer ambiguity

ulanzetz:
Sometimes there is necessary to define TransformerF and Transformer rules that have intersected signatures. Current mechanism of TransformerF try to find both Transformer and TransformerF instances and it's not able to define priority using common Scala implicit priority hacks. It seems logical to me to choose Transformer instance as more priority and don't fail with compilation error on ambiguity. but if it's not possible because it breaks backward compatibility, maybe we can define special flag for this (but I'm not sure about adding this flag to FlagsDsl, because it can be used only for intoF).

Example of ambiguity is presented in tests, Coercible is analog of type class from newtype lib.

WDYT, @krzemin ?

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/TransformerFlags.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/TransformerFlags.scala
@@ -13,4 +13,5 @@ object TransformerFlags {
   final class BeanGetters extends Flag
   final class OptionDefaultsToNone extends Flag
   final class UnsafeOption extends Flag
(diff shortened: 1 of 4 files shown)


Phill101:
@krzemin, it would be so nice to see this feature 🙏

MateuszKubuszok:
Thank you for you contribution! Currently we are working on replacing TransformerF with a build-in structure for validated/failing transformations in https://github.com/scalalandio/chimney/pull/235. You can check the PR and give us an early feedback on its current behavior.

PR #198, opened 2020-12-02
Add a short-circuiting version of TransformerFSupport for Either

mrobakowski:
A short-circuiting version of `TransformerFSupport[Either[E, +*]]` can be very useful and convenient if your code doesn't particularly care what exactly happened and just needs to know if the conversion failed or not. 

I put it in a separate object, so this is a compatible change and requires the user to opt-in via an import.


MateuszKubuszok:
Thank you for you contribution! Currently we are working on replacing `TransformerF` with a build-in structure for validated/failing transformations in https://github.com/scalalandio/chimney/pull/235. It will allow for fail-fast and full-parsing at call-site so a dedicated `TranformerFSupport` will no longer be necessary.

PR #212, opened 2021-10-12
fix: lookup implicit transformers for coproduct members

### LINKED ISSUE #199: bug: failed to transform coproduct with implicit transformer
I get an error with what looks to be a pedestrian case.

```scala
import org.scalatest.funsuite.AnyFunSuite
import org.scalatest.matchers.should.Matchers

sealed trait A extends Product with Serializable
object A {
  case class Foo(s: String) extends A
  case class Bar(kvs: Map[String, Int]) extends A
}

sealed trait B extends Product with Serializable // same structure as A
object B {
  case class Foo(s: String) extends B
  case class Bar(kvs: Map[String, Int]) extends B
}

sealed trait C extends Product with Serializable
object C {
  case class Foo(s: String) extends C
  case class Bar(keys: Seq[String], values: Seq[Int]) extends C
}

class TransformationTest extends AnyFunSuite with Matchers {

  implicit val barTransformer: Transformer[A.Bar, C.Bar] = aBar => {
    val (keys, values) = aBar.kvs.seq.unzip
    C.Bar(keys = keys.toSeq, values = values.toSeq)
  }

  case class Bag[T](xs: Seq[T])

  test("transforming a coproduct with identical structure") {
    val bagA = Bag(Seq(A.Foo("foo"), A.Bar(Map("bar" -> 1))))
    val bagB = bagA.transformInto[Bag[B]]
    succeed
  }

  test("transforming a coproduct with different structure") {
    val bagA = Bag(Seq(A.Foo("foo"), A.Bar(Map("bar" -> 1))))
    val bagC = bagA
      .into[Bag[C]]
      .withFieldComputed(_.xs, _.xs.transformInto[Seq[C]])
      .transform
    succeed
  }

}
```

`A -> B` works, but `A -> C` does not, even if I have a implicit transformer in scope. I get a `Chimney can't derive transformation from Seq[A] to Seq[C]`

image

plaflamme:
I believe this will fix #199 as well as other issues, but somehow it triggers a scalac error that I can't figure out.

This fix is adding an implicit lookup for `Transformer[FromSealedMember, ToSealed]` when deriving `Transformer[FromSealed, ToSealed]` such that individual members can have their own transformer defined. I'm mainly interested in this for dealing with `protobuf` `oneof` (which this does fix), but I believe this will fix other issues (once the scalac issue is figured out).

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
@@ -495,10 +495,17 @@ trait TransformerMacros extends TransformerConfigSupport with MappingMacros with
 
         val instanceClauses = fromInstances.map { instTpe =>
           val instName = instTpe.typeSymbol.name.toString
-
           resolveCoproductInstance(srcPrefixTree, instTpe, To, config)
             .map { instanceTree =>
               Right(cq"_: $instTpe => $instanceTree")
+            }
+            .orElse {
+              resolveImplicitTransformer(config)(instTpe, To)
+                .map { implicitTransformerTree =>
+                  val fn = freshTermName(instName)
+                  val innerTransformerTree = implicitTransformerTree.callTransform(Ident(fn))
+                  Right(cq"$fn: $instTpe => $innerTransformerTree")
+                }
             }
             .getOrElse {
               val instSymbol = instTpe.typeSymbol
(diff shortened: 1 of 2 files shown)


plaflamme:
@krzemin Any insights to provide on this? I believe this will fix a known issue, but parts of this is also beyond my macro-fu, so I would appreciate some guidance if you can provide any.

krzemin:
Thank you @plaflamme for your contribution 🙏 

Followed up in #261, made sure it also works fine with partial transformers.

I resolved the scalac error by just moving the type definitions to a separate top-level object. The compiler sometimes can't handle sealed class hierarchies that are deeply nested.

Issue #214, opened 2022-02-15
Illegal field name


Fristi:
Would this be easy fixable?

Scala 2.13.8
Chimney 0.6.1

I have a case class `ServicepointRowRaw` with a field defined as ```Billing Zip/Postal Code``: String`

```scala
type AntiCorruption[+A] = ValidatedNel[TransformationError[ValidationError], A]

// row: ServicepointRowRaw

row.transformIntoF[AntiCorruption, ServicepointRow]
```

```
java.lang.ClassFormatError: Illegal field name "Billing Zip/Postal Code$macro$16" in class sbs/domain/servicepoints/anticorruption/ServicepointRow$$anon$1
    	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
    	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1012)
    	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150)
    	at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:862)
    	at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:760)
    	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:681)
    	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:639)
    	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
    	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
    	at sbs.domain.servicepoints.anticorruption.ServicepointRow$.$anonfun$transformer$1(ServicepointRow.scala:88)
```

Fristi:
Found the issue

- It seems to occur when you have a case class with a field which contains 2 or more `/` in the field names
- It only occurs when you have to types with the exact same fields name 

This is showcased in #215

krzemin:
fixed in #219

PR #215, opened 2022-02-15
Reproduction for issue 214

Fristi:
Reproduction case for #214


krzemin:
Thank you for the reproduction, used this in #219.

Issue #211, opened 2021-06-14
Patching breaks when using `ignoreRedundantPatcherFields` and ignored fields are not at end of patch object


stephendavidmarsh:
This is best explained with an example:

```scala
import io.scalaland.chimney.dsl.PatcherOps

object {
  case class A(a: String, b: Int)
  case class B(a: String, b: Int, z: Int)
  case class C(z: Int, a: String, b: Int)
  case class D(b: Int, a: String)

  val a = A("a", 5)
  val b = B("b", 6, 7)
  val c = C(9, "c", 8)
  val d = D(10, "d")

  // WORKS
  a.using(b).ignoreRedundantPatcherFields.patch

  // DOESN'T COMPILE
  a.using(c).ignoreRedundantPatcherFields.patch

  // WORKS
  // Error doesn't occur without `ignoreRedundantPatcherFields`
  a.using(d).patch
}
```

Case class C is identical to B except with the unneeded `z` field moved around. Patching with B works, however trying to use C to patch results in this compiler error:

```
error: type mismatch;
        found   : Int
        required: String
```

My best guess is something goes wrong with lining the fields in the two classes up with each other. Rearranging the fields alone isn't enough to trigger the issue, without `ignoreRedundantPatcherFields`, as show with case class D, patching works even if the field order does not match.

an-tex:
Just hit the same. Would be great to get a new release with this bugfix :)

PR #216, opened 2022-04-04
Resolve issue #211

rrramiro:
Resolve issue #211

I've stumble upon the same problem. With the option `ignoreRedundantPatcherFields` patching wasn't working when fields to ignore were at the beginning of the patch case class.


krzemin:
@rrramiro Thank you, your solution looks good 👍

PR #222, opened 2022-07-19
Extract internal *Definition methods to common trait

MateuszKubuszok:
Extract internal `Definition` methods used by macros to a separate trait.


chimney/src/main/scala/io/scalaland/chimney/dsl/TransformerDefinitionCommons.scala
```diff
@@ -0,0 +1,28 @@
+package io.scalaland.chimney.dsl
+
+import io.scalaland.chimney.internal.TransformerCfg
+
+trait TransformerDefinitionCommons[ChangeCfg[_ <: TransformerCfg]] {
```
krzemin, line 5: `UpdateCfg`? so that naming is more aligned with `FlagsDsl`

PR #225, opened 2022-08-02
Refactor: explicit derivation target ADT

krzemin:
This PR brings improved compile-time configuration management using explicit ADT that precisely defines whether which transformer target we're expanding (total `Transformer` or lifted `TransformerF`).

It should not bring any functional changes, but rather open a way for further refactors and code modularization & simplification.


lbialy (approved):
LGTM!

Issue #200, opened 2020-12-28
Default values and type incompatibilites


nilskp:
Motivated by the [issues I've encountered earlier](https://github.com/scalalandio/chimney/issues/176#issuecomment-719835155), and living with `.disableDefaultValues` as a consequence, I think the following rule would be a good balance of convenience and safety: 

> Only apply default values IFF the types still match. 

To recap, my situation was that I had default values defined on a class that I was mapping to, but the type of the source mapping had changed. I was expecting a compiler error, because the types no longer matched, but because the destination had default values, those were silently applied. Very hard to find. Default values are a nice convenience, but not if they end up hiding type incompatibilities.

nilskp:
@krzemin 
Any thoughts on this?

krzemin:
Since 0.7.0 default values processing requires explicit enabling, so it should minimize the consequences in many codebases.

But your point is valid and we may still want to be a bit more restrictive when deciding to apply the default value.

```
case class Foo1(a: Int)
case class Foo2(a: String)
case class Target(a: Int = 1)

implicit val cfg = TransformerConfiguration.default.enableDefaultValues

Foo1(10).transformInto[Target] // should be Target(10)

Foo2("test").transformInto[Target] // should fail the compilation, but succeeds with Target(1)
```

MateuszKubuszok:
Resolved on branch `scala-3` (#325).

Issue #192, opened 2020-10-21
Transform through method other than default constructor


Krever:
Instead of using the default constructor, we could use an alternative method specified by the user. Typically some version of `apply`.

The API could look like `into[Foo].via(Foo.apply).transform` or sth similar.

taeguk:
It will be very useful. Any plans about this? 🙏

mariussoutier:
This would be really nice in order to use a companion object's apply method.

MateuszKubuszok:
Released in 0.8.4

PR #227, opened 2022-08-09
Default values usage require explicit enabling (changed default setting)

MateuszKubuszok:
Due to a lot of issues caused by default values enabled by default (vide https://github.com/scalalandio/chimney/issues/176#issuecomment-719835155, https://github.com/scalalandio/chimney/issues/192, https://github.com/scalalandio/chimney/issues/200), this PR changes the default setting to `.disableDefaultValues`. From now `.enableDefaultValues` should be set explicitly.

This is a breaking change, but one which aims for more safety.

### DIFF
--- a/docs/source/transformers/default-values.rst
+++ b/docs/source/transformers/default-values.rst
@@ -30,17 +30,17 @@ Providing the value manually has always a priority over the default.
   // Butterfly(5, "Steve", "yellow")
 
 
-Disabling default values in generated transformer
+Enabling default values in generated transformer
 -------------------------------------------------
 
-Using ``.disableDefaultValues`` operation it's possible to disable
-lookup for default values and require them always to be passed explicitly.
+Using ``.enableDefaultValues`` operation it's possible to enable
+lookup for default values and fallback to default when there is no transformer
+or config setting the value for field.
 
 .. code-block:: scala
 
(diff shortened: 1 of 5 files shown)


docs/source/transformers/default-values.rst
```diff
@@ -30,17 +30,17 @@ Providing the value manually has always a priority over the default.
   // Butterfly(5, "Steve", "yellow")
 
 
-Disabling default values in generated transformer
+Enabling default values in generated transformer
```
krzemin, line 33: I guess we have to modify the whole page, starting from

> Field's default value is automatically used as a target value when constructing

which no longer holds

docs/source/transformers/default-values.rst
```diff
@@ -1,14 +1,15 @@
 Default values support
 ======================
 
-Chimney respects case classes' default values as a possible target
+Chimney allows case classes' default values as a possible target
 field value source.
 
-Automatic value provision
--------------------------
+Enabling default values in generated transformer
```
krzemin, line 7: I would add warning mentioning that prior to 0.7.0 default values processing was enabled automatically.

MateuszKubuszok: Done

PR #229, opened 2022-11-18
Enable benchmarks on correct runner [CI] 

lbialy:
* Change runs-on to `runs-on: benchmarks`


MateuszKubuszok (approved):
Let's give it a shot

PR #232, opened 2022-12-16
Use supported GH action for Scala rather than deprecated one

### LINKED ISSUE #230: migrate away from olafurpg/setup-scala action in CI
rationale: https://github.com/olafurpg/setup-scala/issues/49

MateuszKubuszok:
Fixes #230


krzemin:
Looks like jvm version as set in matrix is not respected

PR #244, opened 2023-01-24
Refactor tests and increase coverage

MateuszKubuszok:
- [x] split large blocks of tests into smaller organized by funtionality
 - [x] for every flag with add test checking global flag disabling default value and local override
 - [x] make sure that for each transformer type (total, lifted, partial), shared functionality is tested in a similar way to make it easy to look for missing tests

Some work is still in progress but this is mergable to no block the work on [feature/partial-transformers](https://github.com/scalalandio/chimney/tree/feature/partial-transformers) branch


chimney/src/test/scala/io/scalaland/chimney/PartialTransformerProductSpec.scala
```diff
@@ -0,0 +1,1106 @@
...
+        val badTripForm =
+          TripForm("100xyz", List(PersonForm("John", "10", "foo"), PersonForm("Caroline", "bar", "155")))
+
+        val result = badTripForm
+          .intoPartial[Trip]
+          .withFieldComputedPartial(_.id, _.tripId.parseInt.toPartialTransformerResultOrString("bad trip id"))
+          .transform
+
+        result.asOption ==> None
+        result.asEither ==> Left(
```
krzemin, line 908: I'm unsure about constructing these path-prepended error values in tests. They're pretty bad to write and read.

As a middleground, how about using just `asErrorPathMessages` to verify the errors? (look on the rebased branch)

chimney/src/test/scala/io/scalaland/chimney/PartialTransformerResultSpec.scala
```diff
@@ -0,0 +1,54 @@
+package io.scalaland.chimney
+
+import io.scalaland.chimney.dsl._
+import io.scalaland.chimney.examples.trip._
+import utest._
+
+object PartialTransformerResultSpec extends TestSuite {
```
krzemin, line 7: I would expect tests of just `Result` class here (map/flatMap/product/sequence/traverse x short/long circuit), not the PartialTransformer dsl usage...

Maybe also rename to `PartialResultSpec`?

chimney/src/test/scala/io/scalaland/chimney/PartialTransformerSdtLibTypesSpec.scala
```diff
@@ -0,0 +1,428 @@
...
+      case class FooBuzz(unit: Unit)
+      case class ConflictingFooBuzz(value: Unit)
+
+      Buzz("a").transformIntoPartial[NewBuzz].asOption ==> Some(NewBuzz("a", ()))
+      Buzz("a").transformIntoPartial[FooBuzz].asOption ==> Some(FooBuzz(()))
+      NewBuzz("a", null.asInstanceOf[Unit]).transformIntoPartial[FooBuzz].asOption ==> Some(
+        FooBuzz(null.asInstanceOf[Unit])
+      )
+
+      compileError("""Buzz("a").transformIntoPartial[ConflictingFooBuzz]""").check(
```
krzemin, line 28: Check that in a separate (sub)test?

chimney/src/test/scala/io/scalaland/chimney/PartialTransformerSumTypeSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+        .asOption ==> None
+      (short.Trillion("x"): short.NumScale[String, Nothing])
+        .transformIntoPartial[long.NumScale[Int]]
+        .asOption ==> None
+    }
+
+    test(
+      """transforming nested sealed hierarchies from "subset" of case classes to "superset" of case classes without modifiers when common corresponding types are transformable"""
+    ) {
+      // TODO: failing
```
krzemin, line 111: If it's a bug, can you create a ticket and link to it in the TODO comment?

krzemin (changes requested):

chimney/src/test/scala/io/scalaland/chimney/PartialTransformerSumTypeSpec.scala
MateuszKubuszok: It wasn't actually a bug

chimney/src/test/scala/io/scalaland/chimney/PartialTransformerResultSpec.scala
MateuszKubuszok: Renamed, tests are yes to write but tests that used to be there were moved to other suites or got removed if redundant.

krzemin (approved):

PR #251, opened 2023-02-17
Scaladoc documentation added to a few Partial Transformer related classes

MateuszKubuszok:
Added scaladoc where it was skipped


chimney/src/main/scala/io/scalaland/chimney/dsl/ImplicitTransformerPreference.scala
```diff
@@ -1,7 +1,19 @@
...
+  */
 sealed abstract class ImplicitTransformerPreference
 
+/** Tell the derivation to prefer total transformers.
+  *
+  * @since 0.7.0
+  */
 case object PreferTotalTransformer extends ImplicitTransformerPreference
 
+/** Tell the derivation to prefer total transformers.
```
krzemin, line 15: prefer partial

krzemin (approved):

Issue #249, opened 2023-02-11
Improve support for AnyVals


MateuszKubuszok:
Currently we are explicitly supporting:

* `case class Foo(t: T) => T` - unwrapping of type without any conversion
* `T => case class Foo(t: T)` - wrapping if type without any conversion

we should also add:

* `case class Foo(t: T) => case class Bar(t: T)` - re-wrapping of type without any conversion
* `case class Foo(s: S) => T` - unwrapping of type when `S => T` conversion exists
* `S => case class Foo(t: T)` - wrapping if type when `S => T` conversion exists
* `case class Foo(s: S) => case class Bar(t: T)` when `S => T` conversion exists

ekhov:
If you're accepting contributions and this feature is not urgent, I'd be happy to try to implement these :)

krzemin:
The whole scope was addressed in #265 👏

PR #265, opened 2023-03-05
Add rewrapping a value from one value class to another

ekhov:
Implementation of #249

- [x] Add rewrapping value of one value class to another value class if value is the same type.
      ```case class Foo(t: T) extends AnyVal => case class Bar(t: T) extends AnyVal```

- [x] Add unwrapping of a value class's value of one type(`T`) to another type(`S`) if `T => S` transformer exists.
       ```case class Foo(t: T) extends AnyVal => S```
       
- [x] Add wrapping a value of one type(`T`) to value class of another type(`S`) if `T => S` transformer exists.
       ```T => case class Foo(s: S)```
       
- [x]  Add rewrapping value(type `T`) of one value class to another value class(with value of type `S`  if `T => S` transformer exists.
        ```case class Foo(t: T) => case class Bar(s: S)```

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
@@ -169,7 +169,7 @@ trait TransformerMacros extends MappingMacros with TargetConstructorMacros with 
           expandSubtypes(srcPrefixTree, config)
         } else if (bothValueClasses(From, To)) {
           expandValueClassToValueClass(srcPrefixTree, config)(From, To)
-        } else if (fromValueClassToType(From, To)) {
+        } else if (fromValueClass(From, To)) {
           expandValueClassToType(srcPrefixTree, config)(From, To)
         } else if (fromTypeToValueClass(From, To)) {
           expandTypeToValueClass(srcPrefixTree, config)(From, To)
@@ -241,21 +241,19 @@ trait TransformerMacros extends MappingMacros with TargetConstructorMacros with 
       To: Type
   ): Either[Seq[TransformerDerivationError], Tree] = {
 
-    From.valueClassMember
-      .map { member =>
-        Right {
-          mkTransformerBodyTree0(config.derivationTarget) {
-            q"$srcPrefixTree.${member.name}"
-          }
-        }
-      }
-      .getOrElse {
-        // $COVERAGE-OFF$
-        Left {
-          Seq(CantFindValueClassMember(From.typeSymbol.name.toString, To.typeSymbol.name.toString))
-        }
-        // $COVERAGE-ON$
-      }
+    val fromValueClassMember = From.valueClassMember.toRight(
+      Seq(CantFindValueClassMember(From.typeSymbol.name.toString, To.typeSymbol.name.toString))
+    )
+
+    for {
+      fromValueClassMember <- fromValueClassMember
+      fromValueClassMemberType = fromValueClassMember.returnType
+      fromMemberAccessTree = q"$srcPrefixTree.${fromValueClassMember.name}"
+      transformerBodyTree <- resolveRecursiveTransformerBody(fromMemberAccessTree, config)(
+        fromValueClassMemberType,
+        To
+      )
+    } yield mkTransformerBodyTree0(transformerBodyTree.target)(transformerBodyTree.tree)
   }
 
   def expandTypeToValueClass(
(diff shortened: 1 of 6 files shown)


krzemin:
Thanks @ekhov, looks good so far 👍 

In order to cover the following cases, I guess the easiest way would be to generalize existing value class rules (including the one you have created), so that they work with any recursively derived `S => T` transformation (which should cover also the specific `T => T` case).

`resolveRecursiveTransformerBody` will be your friend.

chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
```diff
@@ -239,21 +241,19 @@ trait TransformerMacros extends MappingMacros with TargetConstructorMacros with
...
+
+    for {
+      fromValueClassMember <- fromValueClassMember
+      fromValueClassMemberType = fromValueClassMember.returnType
+      fromMemberAccessTree = q"$srcPrefixTree.${fromValueClassMember.name}"
+      transformerBodyTree <- resolveRecursiveTransformerBody(fromMemberAccessTree, config)(
+        fromValueClassMemberType,
+        To
+      )
+    } yield mkTransformerBodyTree0(transformerBodyTree.target)(transformerBodyTree.tree)
```
krzemin, line 256: You should use `config.derivationTarget` - this is the most probable reason why the tests are failing.
Same applies to the remaining rules.

krzemin, line 256:
```suggestion
    } yield transformerBodyTree.tree
```
Returning just recursively derived tree should be sufficient in this case.

chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
```diff
@@ -263,11 +263,18 @@ trait TransformerMacros extends MappingMacros with TargetConstructorMacros with
...
+    )
+
+    for {
+      toValueClassMemberType <- toValueClassMember.map(_.returnType)
+      transformerBodyTree <- resolveRecursiveTransformerBody(srcPrefixTree, config)(
+        From,
+        toValueClassMemberType
+      )
+      toCreationTree = q"new $To(${transformerBodyTree.tree})"
+    } yield mkTransformerBodyTree0(transformerBodyTree.target)(toCreationTree)
```
krzemin, line 277:
```suggestion
    for {
      toValueClassMethodSymbol <- toValueClassMember
      toValueClassMemberType <- toValueClassMember.map(_.returnType)
      transformerBodyTree <- resolveRecursiveTransformerBody(srcPrefixTree, config)(
        From,
        toValueClassMemberType
      )
    } yield mkTransformerBodyTree1(
      To,
      Target.fromField(toValueClassMethodSymbol, From),
      transformerBodyTree,
      config.derivationTarget
    )(innerTree => q"new $To(${innerTree})")
```
This one is going to be quite more complicated to get it right.
It's not sufficient to just wrap recursively derived tree in `new To(...)`, as recursive tree might be lifted/partial transformer that you have to map over. For such cases `mkTransformerBodyTree1` should be used as in the snipped.

chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
```diff
@@ -282,6 +289,35 @@ trait TransformerMacros extends MappingMacros with TargetConstructorMacros with
...
+      fromValueClassMember <- fromValueClassMember
+      fromValueClassMemberType = fromValueClassMember.returnType
+      toValueClassMemberType <- toValueClassMember.map(_.returnType)
+      fromMemberAccessTree = q"$srcPrefixTree.${fromValueClassMember.name}"
+      transformerBodyTree <- resolveRecursiveTransformerBody(fromMemberAccessTree, config)(
+        fromValueClassMemberType,
+        toValueClassMemberType
+      )
+      toCreationTree = q"new $To(${transformerBodyTree.tree})"
+    } yield mkTransformerBodyTree0(transformerBodyTree.target)(toCreationTree)
```
krzemin, line 318: This rule may require similar `mkTransformerBodyTree1` adjustment as in `expandTypeToValueClass` rule.
Please add at least at test that checks transformation between two value classes, relying on implicit partial transformer to transform their inner types.

krzemin (changes requested):
Hi, thanks for your submission! I left few comments with suggestions how to fix failing unit tests.

chimney/src/main/scala/io/scalaland/chimney/internal/macros/TransformerMacros.scala
ekhov: @krzemin Thanks for the suggestions! Sorry for not delivering for too long- I managed to figure these ones myself, but didn't have much time to actually make a fix. But I'm just working on the fix right now, so hope today PR will be ready!

krzemin (approved):
LGTM 👍

Issue #279, opened 2023-03-10
Use typed abstractions in macros


MateuszKubuszok:
Currently we use a lot of `c.universe.Type` and `c.universe.Tree` inside macros which would make cross-compilation with 3 hard.

For shared code, we create abstract types `Type[T]` and `Expr[T]` and use them for passing data around [like here](https://github.com/MateuszKubuszok/pipez/blob/master/pipez/src/main/scala/pipez/internal/Definitions.scala#L21).

But a prerequisite for that is using typed abstractions instead of untyped in the first place. We could define them [like here](https://github.com/MateuszKubuszok/pipez/blob/master/pipez/src/main/scala-2/pipez/internal/PlatformDefinitions.scala#L18).

This would be helpful in #276 .

krzemin:
Cross-compilable typed abstractions were delivered in #286.

Issue #276, opened 2023-03-07
Refactor `TransformerMacros` into rule-based, cross-compilable architecture


krzemin:
Scala 2 chimney dispatches transformation rules as a single if-then-else block inside `TransformerMacros` module.

In the scope of this task we should:
- define an interface that can express transformation rules
- rewrite existing transformation logic in terms of independent rules which should end up in separate files
- make sure the rules can call derivation logic recursively
- new logic should call cross-compilable metaprogramming abstractions so that they generate semantically-equivalent code for both Scala 2 and 3

Rules to cover (names following existing Scala 2 impl):
- [x] `expandSubtypes` - implemented in https://github.com/scalalandio/chimney/pull/296 to verify that it works
- [x] `expandOptions` - https://github.com/scalalandio/chimney/issues/303
- [x] `expandPartialFromOptionToNonOption` - https://github.com/scalalandio/chimney/issues/305
- [x] `expandTargetWrappedInOption` - https://github.com/scalalandio/chimney/issues/305
- [x] `expandSourceWrappedInOption` - https://github.com/scalalandio/chimney/issues/306
- [x] `expandValueClassToValueClass` - https://github.com/scalalandio/chimney/issues/307
- [x] `expandValueClassToType` - https://github.com/scalalandio/chimney/issues/308
- [x] `expandTypeToValueClass` - https://github.com/scalalandio/chimney/issues/309
- [x] `expandEithers` - https://github.com/scalalandio/chimney/issues/310
- [x] `expandFromMap` - https://github.com/scalalandio/chimney/issues/311
- [x] `expandIterableOrArray` - https://github.com/scalalandio/chimney/issues/312
- [x] `expandDestinationTuple` - https://github.com/scalalandio/chimney/issues/313
- [x] `expandDestinationCaseClass` - https://github.com/scalalandio/chimney/issues/314
- [x] `expandDestinationJavaBean` - https://github.com/scalalandio/chimney/issues/315
- [x] `expandSealedClasses` - https://github.com/scalalandio/chimney/issues/316

MateuszKubuszok:
Merged to `scala-3` in #320.

PR #286, opened 2023-03-15
Cross compilable Scala 2/3 abstractions

MateuszKubuszok:
This PR is a draft proposal for an alternative architecture of macros. (Touching #279 #276).

TODO:
 - [x] create an entrypoint to call this new structore from Scala 2 macros
 - [x] create a temporary bridge which would delegate not-yet-implemented branched to old macro code
 - [x] create an initial implementation for this structure in Scala 3 to show that it actually does enable implementation
 - [ ] make Scala 3 version return `???` as trees to make it compile all cases but fail tests in runtime to make the migration process less PITA


chimney/src/main/scala-3/io/scalaland/chimney/compiletime/DefinitionsPlatform.scala
```diff
@@ -0,0 +1,150 @@
+package io.scalaland.chimney.internal.compiletime
+
+import io.scalaland.chimney.dsl as dsls
+import io.scalaland.chimney.internal
+import io.scalaland.chimney.{partial, PartialTransformer, Patcher, Transformer}
+
+import scala.quoted
+
+private[compiletime] trait DefinitionsPlatform(using val quotes: quoted.Quotes)
```
krzemin, line 9: `using quotes: quoted.Quotes` not enough? why do you need val?

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/ResultDefinitions.scala
```diff
@@ -0,0 +1,324 @@
...
+
+  /** Contains everything we would like to access in our computations without resorting to globals.
+   *
+   * Possible usages:
+   * - access to config
+   * - appending logs and diagnostics
+   * - storing cache for intermediate results
+   */
+  final case class Context(
+      config: TransformerConfig = TransformerConfig(),
```
krzemin, line 44: I'd avoid coupling result monad with just transformer-specific stuff. In the end we may use it for derive patchers too. Wouldn't it be too hard to parameterize the context at this point?

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/ResultDefinitions.scala
```diff
@@ -0,0 +1,324 @@
...
+     *   andThen
+     *   C => F[D])
+     * )
+     * ...
+     * }}}
+     *
+     * which would make it possible to evaluate the function at the head. By rewriting until evaluation is possible,
+     * and then evaluating in a `loop`, we are able to execute the whole `Result` with stack safety.
+     */
+    sealed private trait Stack[-I, O]
```
krzemin, line 275: holy mother 🙄

krzemin (changes requested):

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Definitions.scala
krzemin, line 1: How about keeping types and exprs in separate modules?

Maybe even basic/stdlib ones should be separated from library-specific definitions?

chimney/src/main/scala-2/io/scalaland/chimney/internal/compiletime/ConfigurationsPlatform.scala
```diff
@@ -0,0 +1,169 @@
...
+    // TODO: adjust for new config type
+    // TODO: this coule be tailrec
+    private def extractTransformerConfig[From: Type, To: Type, Cfg <: internal.TransformerCfg: Type](
+        runtimeDataIdx: Int
+    ): TransformerConfig[From, To] = {
+      /*
+      val cfgTpe = Type[Cfg].dealias
+
+      if (cfgTpe =:= emptyT) {
+        TransformerConfig()
```
MateuszKubuszok, line 115: This needs to be updated to:
 * add right _typed_ `Field` or `CoproductOverride`
 * add untyped coproduct override in `config.legacy`

chimney/src/main/scala-2/io/scalaland/chimney/internal/compiletime/ResultsPlatform.scala
```diff
@@ -0,0 +1,7 @@
+package io.scalaland.chimney.internal.compiletime
+
+private[compiletime] trait ResultsPlatform extends Results { this: DefinitionsPlatform =>
+
+  // TODO: proper error generation
+  protected def reportError(errors: String): Nothing = c.abort(c.enclosingPosition, errors)
```
MateuszKubuszok, line 6: At some point, somewhere, we will have to turn the collection of errors into `String` to display, but we can do it at the very end.

chimney/src/main/scala-2/io/scalaland/chimney/internal/compiletime/derivation/LegacyPlatform.scala
```diff
@@ -0,0 +1,41 @@
...
+private[compiletime] trait LegacyPlatform extends Legacy {
+  this: DefinitionsPlatform & ConfigurationsPlatform & Contexts =>
+
+  protected object legacy extends LegacyImpl {
+
+    private def convertToLegacyConfig[From, To](
+        tbm: TransformerBlackboxMacros
+    )(implicit ctx: Context.ForTransformer[From, To]): tbm.TransformerConfig = {
+      // TODO: convert our new config into our old config
+      ???
```
MateuszKubuszok, line 16: I intend this to use values from Context (including `ctx.config.legacy`) to generate `TransformerConfig` and pass it to original macro generating code.

chimney/src/main/scala-2/io/scalaland/chimney/internal/macros/TransformerMacros.scala
```diff
@@ -24,6 +24,12 @@ trait TransformerMacros extends MappingMacros with TargetConstructorMacros with
       .withDefinitionScope(weakTypeOf[From], weakTypeOf[To])
       .withDerivationTarget(derivationTarget)
 
+    buildDefinedTransformerFromConfig[From, To](config)
+  }
+
+  def buildDefinedTransformerFromConfig[From: WeakTypeTag, To: WeakTypeTag](
+      config: TransformerConfig
+  ): Tree = {
```
MateuszKubuszok, line 32: Simple split of a method to make calling from outside easier.

chimney/src/main/scala-3/io/scalaland/chimney/compiletime/DefinitionsPlatform.scala
MateuszKubuszok: Without `val` you would not see this `quotes` in mixins and you would not be able to import it.

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Configurations.scala
```diff
@@ -0,0 +1,149 @@
+package io.scalaland.chimney.internal.compiletime
+
+import io.scalaland.chimney.dsl.ImplicitTransformerPreference
+import io.scalaland.chimney.internal
+import io.scalaland.chimney.partial
+import io.scalaland.chimney.dsl as dsls
+
+import scala.annotation.nowarn
+
+@nowarn("msg=The outer reference in this type test cannot be checked at run time.")
```
krzemin, line 10: Any idea why it is raised? Maybe add a TODO comment so that we don't forget to go back to it before release?

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Configurations.scala
```diff
@@ -0,0 +1,149 @@
...
+      legacy: TransformerConfig.LegacyData = TransformerConfig.LegacyData() // TODO: temporary
+  )
+  object TransformerConfig {
+
+    // TODO: for creating TransformerConfig for old macros in Scala 2 until everything is migrated
+    final case class LegacyData(
+        fieldOverrideLegacy: Map[String, FieldOverride.RuntimeConfiguration] = Map.empty,
+        coproductInstanceOverridesLegacy: Map[(ComputedType, ComputedType), Int] = Map.empty,
+        coproductInstancesPartialOverridesLegacy: Map[(ComputedType, ComputedType), Int] = Map.empty,
+        transformerDefinitionPrefix: Expr[dsls.TransformerDefinitionCommons.RuntimeDataStore] =
```
krzemin, line 128: In current master we have:

```
      transformerDefinitionPrefix: Tree = EmptyTree,
      definitionScope: Option[(Type, Type)] = None
```

We're probably going to need those in order to preserve current behavior. And not talking only about legacy config, but also in the new one.

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Contexts.scala
```diff
@@ -0,0 +1,132 @@
...
+  final protected class LogEntry(val nesting: Int, thunk: => String) {
+    lazy val message: String = thunk
+  }
+
+  sealed protected trait Context {
+
+    type Target
+    type Typeclass
+
+    def logs: Vector[LogEntry]
```
krzemin, line 21: perhaps better to move the logging capability to DerivationResult

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Contexts.scala
```diff
@@ -0,0 +1,132 @@
...
+        To = implicitly[Type[To]],
+        src = src,
+        failFast = failFast,
+        config = config,
+        logs = Vector.empty,
+        logNesting = 0
+      )
+    }
+
+    final case class ForPatcher[T, Patch](
```
krzemin, line 95: Do you see any benefits from shared context base type for patchers and transformers?

My intuition is that the derivation code for both TCs will be quote independent (perhaps one calling other, but that would require building context from scratch anyway).

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Configurations.scala
MateuszKubuszok: For Scala compiler _every_ `trait`, `abstract class` and `case class` which is not `private` is treated as potentially leaking, and accessible from e.g. other instances of our macro class. As such type tests would not catch that this is e.g. `macro2.DerivationResult` and not `macro1.DerivationResult` and compiler complains about it.

I tried using `protected[this]` in attempt to make it visible only to current instance but:
* it didn't restrict access at all, warning remained
* `protected[this]` is deprecated in Scala :P

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Configurations.scala
MateuszKubuszok: They are already stored in `Context[From, To]` and function converting to legacy config takes a whole context.

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Contexts.scala
MateuszKubuszok: Patchers are already pulling in Transformers so they would use potentially all of transformers functionality. They would surely share:
 * Types
 * Exprs
 * Results

Then we can theoretically turn `Definitions` into `TransformerDefinitions` and exclude `Patcher`s but it would inevitable require us to do:

```scala
trait TransformerDefinitions
  extends Types
  with Exprs
  with TransformerConfigurations
  with TransformerContexts
  with Results {
 ...
}

trait PatcherDefinitions
  extends TransformerDefinitions
  with PatcherConfigurations
  with PatcherContexts
```

and I am not sure there would be many benefits in such granularity.

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Exprs.scala
```diff
@@ -0,0 +1,145 @@
+package io.scalaland.chimney.internal.compiletime
+
+import io.scalaland.chimney.partial
+
+private[compiletime] trait Exprs { this: Definitions =>
+
+  /** Platform-specific expression representation (c.universe.Expr[A] in 2, quotes.Expr[A] in 3 */
+  protected type Expr[A]
+  protected object Expr {
```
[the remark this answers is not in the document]
MateuszKubuszok: Scope pollution. Previous version with all methods directly under this give you 40+ methods not grouped by any functionality in current scope. And you cannot make it `private` because you have to implement it in a platform-dependent mixin.

Meanwhile moving all extensions to `MethodsImpl` and all implementations to some `CompanionImpl` works relatively nice to API users, it's approach used in Scala 3 `Quotes`.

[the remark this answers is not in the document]
krzemin: Instead of the following:
```scala
// Expr
object Option {
  def apply[A: Type](a: Expr[A]): Expr[Option[A]] = exprImpl.Option[A](a)
  def empty[A: Type]: Expr[Option[A]] = exprImpl.OptionEmpty[A]
  def apply[A: Type]: Expr[A => Option[A]] = exprImpl.OptionApply[A]
}
// ExprImpl
def Option[A: Type](value: Expr[A]): Expr[Option[A]]
def OptionEmpty[A: Type]: Expr[Option[A]]
def OptionApply[A: Type]: Expr[A => Option[A]]
// implementations of ExprImpl in platform-specific code
```

Why can't you have just the following?
```scala
def Option: OptionModule
trait OptionModule {
  def apply[A: Type](a: Expr[A]): Expr[Option[A]]
  def empty[A: Type]: Expr[Option[A]]
  def apply[A: Type]: Expr[A => Option[A]]
}
// concrete implementation of Option trait in platform-specific code
```

This way you should be able to keep the clean scope (even multiply-nested) and avoid maintaining approximately half of the code of the facade. This should be also closer to what you have in Scala 3 Quotes.

chimney/src/main/scala-2/io/scalaland/chimney/internal/compiletime/DefinitionsPlatform.scala
```diff
@@ -0,0 +1,17 @@
...
+    with TypesPlatform
+    with ExprsPlatform
+    with ConfigurationsPlatform
+    with ResultsPlatform {
+
+  val c: blackbox.Context
+}
+private[compiletime] object DefinitionsPlatform {
+  type Arbitrary
+  type Arbitrary2
```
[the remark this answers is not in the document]
krzemin: Looks like they can be replaced by wildcard types (`?`)

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Exprs.scala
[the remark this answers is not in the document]
krzemin: Addressed in #293

chimney/src/main/scala-2/io/scalaland/chimney/internal/compiletime/DefinitionsPlatform.scala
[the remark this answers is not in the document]
krzemin: Addressed in #293

chimney/src/main/scala-2/io/scalaland/chimney/internal/compiletime/ConfigurationsPlatform.scala
krzemin: it's about to be handled in a separate PR, as agreed

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Configurations.scala
krzemin: they were not immediately, but they were added in the process... 👍

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/DerivationDefinitions.scala
```diff
@@ -0,0 +1,91 @@
+package io.scalaland.chimney.internal.compiletime.derivation
+
+import io.scalaland.chimney.internal.compiletime.Definitions
+import io.scalaland.chimney.{partial, PartialTransformer, Transformer}
+
+import scala.annotation.nowarn
+
+@nowarn("msg=The outer reference in this type test cannot be checked at run time.")
+private[derivation] trait DerivationDefinitions { this: Definitions =>
```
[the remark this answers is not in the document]
krzemin: moved to a separate package

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/DerivationGateway.scala
```diff
@@ -0,0 +1,62 @@
+package io.scalaland.chimney.internal.compiletime.derivation
+
+import io.scalaland.chimney.{PartialTransformer, Transformer}
+import io.scalaland.chimney.internal
+import io.scalaland.chimney.internal.compiletime.Definitions
+import io.scalaland.chimney.partial
+
+// TODO
+private[compiletime] trait DerivationGateway { this: Definitions & DerivationDefinitions =>
```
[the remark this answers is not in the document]
krzemin: moved to a separate package

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/ConfigurationDefinitions.scala
```diff
@@ -0,0 +1,144 @@
+package io.scalaland.chimney.internal.compiletime
+
+import io.scalaland.chimney.dsl.ImplicitTransformerPreference
+import io.scalaland.chimney.internal
+
+import scala.annotation.nowarn
+
+@nowarn("msg=The outer reference in this type test cannot be checked at run time.")
+private[compiletime] trait ConfigurationDefinitions { this: Definitions =>
```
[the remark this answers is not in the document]
krzemin: moved to a separate package

PR #294, opened 2023-04-18
Share TransformerDefinitionErrors, split TransformerContext and PatcherContext, fix legacy macro workaround implementation

MateuszKubuszok:
Addresses several points from the discussion in https://github.com/scalalandio/chimney/pull/286:
 * Contexts for Transformers and Patchers are split
 * errors ADT we are using in macros are shared with old macros, to make transition easier
 * correct which routine should be used for recursive derivation and how we handle transformed value Expr (distinction between Total Transformer result and Partial Transformer result):
   * both when we call the old macros for the transition period
   * the way we would model thing in new macros


chimney/src/main/scala/io/scalaland/chimney/internal/TransformerDerivationError.scala
```diff
@@ -47,6 +48,12 @@ final case class IncompatibleSourceTuple(
 final case class NotSupportedTransformerDerivation(fieldName: String, sourceTypeName: String, targetTypeName: String)
     extends TransformerDerivationError
 
+final case class MacroException(exception: Throwable, sourceTypeName: String, targetTypeName: String)
```
krzemin, line 51: `MacroException`/`NotYetImplemented` don't seem to be transformer-specific errors. How about extracting them to a separate hierarchy?

krzemin, line 51: `MacroException`/`NotYetImplemented` don't seem to be transformer-specific errors. How about extracting them to a separate hierarchy?

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Results.scala
```diff
@@ -1,12 +1,35 @@
...
+
 import scala.annotation.nowarn
 import scala.collection.compat.*
 import scala.util.control.NonFatal
 
 @nowarn("msg=The outer reference in this type test cannot be checked at run time.")
 private[compiletime] trait Results { this: Definitions =>
 
+  // TODO: consider expanding
+  trait Trace {
```
krzemin, line 13: Looks like something transformer-specific. Is it possible to decouple it from Result?

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/Gateway.scala
```diff
@@ -4,14 +4,17 @@ import io.scalaland.chimney.{PartialTransformer, Transformer}
...
 private[compiletime] trait Gateway { this: Definitions & Derivation & Legacy =>
 
-  /** Intended for: being called from *Unsafe method to return final Expr; recursive derivation */
+  /** Intended for: being called from *Unsafe method to return final Expr */
   final protected def deriveTotalTransformer[From: Type, To: Type](
       config: TransformerConfig[From, To]
   ): DerivationResult[Expr[Transformer[From, To]]] =
     instantiateTotalTransformer[From, To] { (src: Expr[From]) =>
-      deriveTransformerBody(Context.ForTotal.create[From, To](src = src, config))
+      deriveTransformerTargetExpr(TransformerContext.ForTotal.create[From, To](src = src, config))
```
krzemin, line 17: just `src` should be fine

chimney/src/main/scala/io/scalaland/chimney/internal/TransformerDerivationError.scala
krzemin: duplicate of https://github.com/scalalandio/chimney/pull/294#discussion_r1170509241

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/Gateway.scala
```diff
@@ -46,9 +49,27 @@ private[compiletime] trait Gateway { this: Definitions & Derivation & Legacy =>
...
+      case DerivedExpr.PartialExpr(expr) =>
+        ctx match {
+          case TransformerContext.ForTotal(_, _, _, _) =>
+            DerivationResult.fromException(
+              new AssertionError("Derived partial.Result expression where total Transformer excepts direct value")
+            )
+          case TransformerContext.ForPartial(_, _, _, _, _) => DerivationResult.pure(expr)
+        }
+    }
     DerivationResult.notYetImplemented("Actual derivation")
```
krzemin, line 73: it's now implemented

krzemin (approved):

PR #296, opened 2023-04-20
Kick off rule-based architecture in new macros

MateuszKubuszok:
Prerequisite for https://github.com/scalalandio/chimney/issues/276

- [x] create `Rule` abstraction where each derivation rule would be implemented with this interface
- [x] create method for attempting to use all listed rules and reporting if all of them failed
- [x] define (platform-specific) `rules` list where all rules implemented for each platform could be added
- [x] add fallback to old macros in Scala 2 rules (temporarily, until all rules are migrated)
- [x] add fallback to `???` implementation in Scala 3 rules (temporarily, until all rules are migrated)
- [x] use Rules in `Derivation`
- [x] pretty print `Type`s and `Expr`s (to allow printing correct error when no rule matched)
- [x] add "into" variants of macros to `Gateway` (TransformerInto, PartialTransformerInto)
- [x] create `Macro` class/object in platform specific code directly usable from platform specific macros
- [x] parse config passed as exprs/types into new config format
- [x] start calling new macros from Scala 2 DSL and Scala 3 DSL
  - [x] derive Transformer
  - [x] transform into
  - [x] derive PartialTransformer
  - [x] transform into partial
- [x] implement first rule (subtyping) to test whether generator switched to new architecture properly


chimney/src/main/scala-2/io/scalaland/chimney/internal/compiletime/ConfigurationsPlatform.scala
```diff
@@ -104,65 +86,81 @@ private[compiletime] trait ConfigurationsPlatform extends Configurations { this:
 
     // TODO: adjust for new config type
```
krzemin, line 87: Are these TODO comments still relevant?

For tailrec I believe we agreed to leave it for now, but how about the other one?

MateuszKubuszok: I think this TODO was overlooked during cleanup, it can be removed

krzemin: removed ✅

krzemin (approved):
LGTM

PR #299, opened 2023-04-28
Move to new macros endpoints

MateuszKubuszok:
PR to help keeping trace of endpoints rewired:

 * Scala 2 macros should move new endpoints (they call old macros as a fallback)
 * Scala 3 macros should start using a new macros in the first place
 * found bugs (mostly in Scala 3) should be fixed so that CI errors would be only due to not yet ported rules

- [x] `Transformer.derive`
  - [x] Scala 2 DSL
  - [x] Scala 3 DSL
- [x] `PartialTransformer.derive`
  - [x] Scala 2 DSL
  - [x] Scala 3 DSL
- [x] `Transformer.define.buildTransformer`
  - [x] Scala 2 DSL
  - [x] Scala 3 DSL
- [x] `PartialTransformer.define.buildTransformer` 
  - [x] Scala 2 DSL
  - [x] Scala 3 DSL
- [x] `.transformInto`
  - [x] Scala 2 DSL
  - [x] Scala 3 DSL
- [x] `.transformIntoPartial`
  - [x] Scala 2 DSL
  - [x] Scala 3 DSL
- [x] `.into.transform`
  - [x] Scala 2 DSL
  - [x] Scala 3 DSL
- [x] `.intoPartial.transform`
  - [x] Scala 2 DSL
  - [x] Scala 3 DSL
- [x] `.intoPartial.transformFailFast`
  - [x] Scala 2 DSL
  - [x] Scala 3 DSL
 
- [x] fix errors with Scala 3 TransformerCfg parsing
- [x] implement String singleton type to String parsing in config reading

### DIFF
--- a/chimney/src/main/scala-3/io/scalaland/chimney/internal/compiletime/derivation/transformer/TransformerMacros.scala
+++ b/chimney/src/main/scala-3/io/scalaland/chimney/internal/compiletime/derivation/transformer/TransformerMacros.scala
@@ -103,6 +103,18 @@ object TransformerMacros {
   )(using quotes: Quotes): Expr[Transformer[From, To]] =
     new TransformerMacros(quotes).deriveTotalTransformerWithConfig[From, To, Cfg, Flags, ScopeFlags](td)
 
+  final def deriveTotalTransformerResultWithConfig[
+      From: Type,
+      To: Type,
+      Cfg <: internal.TransformerCfg: Type,
+      Flags <: internal.TransformerFlags: Type,
+      ScopeFlags <: internal.TransformerFlags: Type
+  ](source: Expr[From], td: Expr[TransformerDefinition[From, To, Cfg, Flags]])(using quotes: Quotes): Expr[To] =
+    new TransformerMacros(quotes).deriveTotalTransformationResult[From, To, Cfg, Flags, ScopeFlags](
+      source,
+      Some('{ ${ td }.runtimeData })
+    )
+
   final def derivePartialTransformerWithDefaults[
       From: Type,
       To: Type
--- a/chimney/src/main/scala-3/io/scalaland/chimney/internal/compiletime/derivation/transformer/TransformerMacros.scala
+++ b/chimney/src/main/scala-3/io/scalaland/chimney/internal/compiletime/derivation/transformer/TransformerMacros.scala
@@ -20,7 +20,9 @@ final class TransformerMacros(q: Quotes)
       To: Type
   ](using quotes: Quotes): Expr[Transformer[From, To]] =
     resolveImplicitScopeFlagsAndMuteUnusedConfigWarnings { implicit ImplicitScopeFlagsType =>
-      deriveTotalTransformer[From, To, Empty, Default, ImplicitScopeFlagsType](runtimeDataStore = None)
+      deriveTotalTransformer[From, To, Empty, Default, ImplicitScopeFlagsType](
+        runtimeDataStore = ChimneyExpr.RuntimeDataStore.empty
+      )
     }
 
   final def deriveTotalTransformerWithConfig[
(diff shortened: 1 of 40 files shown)


chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/Gateway.scala
```diff
@@ -99,18 +107,18 @@ private[compiletime] trait Gateway { this: Definitions & Derivation =>
       .flatMap {
         case DerivedExpr.TotalExpr(expr) =>
           ctx match {
-            case TransformerContext.ForTotal(_, _, _, _) =>
+            case TransformerContext.ForTotal(_, _, _, _, _) =>
```
krzemin, line 110: I'd stick to type-based matching in such cases (`case _: TransformerContext.ForTotal`).

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Contexts.scala
```diff
@@ -42,11 +44,22 @@ private[compiletime] trait Contexts { this: Definitions & Configurations =>
...
+          src: Expr[From],
+          config: TransformerConfig,
+          runtimeDataStore: Option[Expr[TransformerDefinitionCommons.RuntimeDataStore]]
+      ): ForTotal[From, To] =
         ForTotal(
           From = Type[From],
           To = Type[To],
           src = src,
+          runtimeDataStore = runtimeDataStore
+            .getOrElse(
```
krzemin, line 57: There's an inconsistency between definition of context (where `runtimeDataStore` expr is required) and factory method (when it's actually optional). I'd strive for unifying so that we don't have to construct dummy expressions that accessing might be unsafe later.

MateuszKubuszok: I was thinking about using temporary hack with something like:
```
  case q"""${something}.runtimeData""" => 
```
in legacy macros rule, to avoid big changes.

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Contexts.scala
```diff
@@ -75,12 +89,19 @@ private[compiletime] trait Contexts { this: Definitions & Configurations =>
...
         From = Type[From],
         To = Type[To],
         src = src,
         failFast = failFast,
+        runtimeDataStore = runtimeDataStore.getOrElse(
+          Expr.asInstanceOf[Nothing, TransformerDefinitionCommons.RuntimeDataStore](Expr.Nothing)(
+            Type.Nothing,
+            ChimneyType.RuntimeDataStore
+          )
+        ),
```
MateuszKubuszok, line 104: Here, I am also on the edge. I would like to prevent setting up some stubs when we aren't using `RuntimeDataStore` but at the same time I want to be able to assume that when we're using it, we are not shoving `Option`s  back and forth - I guess when we'll start using it directly we can try out a few things and see what works best (`null` as default, assert != null on first use, `???` expr as default which would not be used etc).

chimney/src/main/scala-3/io/scalaland/chimney/internal/compiletime/derivation/transformer/TransformerMacros.scala
```diff
@@ -10,33 +13,45 @@ final class TransformerMacros(q: Quotes)
     with DerivationPlatform
     with GatewayPlatform {
 
+  type LocalConfigType <: internal.TransformerFlags
```
krzemin, line 16: naming: it doesn't seem to be a config type, but flags type

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Contexts.scala
krzemin: We can always expr-ize `TransformerDefinitionCommons.emptyRuntimeDataStore` as empty collection expr and keep the context simple, without options.

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/Contexts.scala
krzemin: See https://github.com/scalalandio/chimney/pull/299#discussion_r1183601495

PR #317, opened 2023-05-04
Logging diagnostics in macros

MateuszKubuszok:
- [x] write code logging messages from `DerivationResult`, derivation time and final expression
- [x] add flag which enables logging diagnostic messages during derivation
- [x] add logs for current derivation rules

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/dsl/FlagsDsl.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/dsl/FlagsDsl.scala
@@ -133,6 +133,22 @@ private[dsl] trait FlagsDsl[UpdateFlag[_ <: TransformerFlags], Flags <: Transfor
   def disableImplicitConflictResolution: UpdateFlag[Disable[ImplicitConflictResolution[?], Flags]] =
     disableFlag[ImplicitConflictResolution[?]]
 
+  /** Enable printing the logs from the derivation process.
+   *
+   * @see [[https://scalalandio.github.io/chimney/TODO]] for more details
+   * @since 0.8.0
+   */
+  def enableDerivationLog: UpdateFlag[Enable[DerivationLog, Flags]] =
+    enableFlag[DerivationLog]
+
+  /** Disable printing the logs from the derivation process.
+   *
+   * @see [[https://scalalandio.github.io/chimney/TODO]] for more details
+   * @since 0.8.0
+   */
+  def disableDerivationLog: UpdateFlag[Disable[DerivationLog, Flags]] =
+    disableFlag[DerivationLog]
+
   private def enableFlag[F <: TransformerFlags.Flag]: UpdateFlag[Enable[F, Flags]] =
     this.asInstanceOf[UpdateFlag[Enable[F, Flags]]]
 
--- a/chimney/src/main/scala/io/scalaland/chimney/dsl/FlagsDsl.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/dsl/FlagsDsl.scala
@@ -138,16 +138,16 @@ private[dsl] trait FlagsDsl[UpdateFlag[_ <: TransformerFlags], Flags <: Transfor
    * @see [[https://scalalandio.github.io/chimney/TODO]] for more details
    * @since 0.8.0
    */
-  def enableDerivationLog: UpdateFlag[Enable[DerivationLog, Flags]] =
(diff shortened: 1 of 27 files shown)


chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformSubtypesRuleModule.scala
```diff
@@ -8,17 +8,10 @@ trait TransformSubtypesRuleModule { this: Definitions & Derivation =>
...
 
-    def apply[From, To](implicit ctx: TransformerContext[From, To]): DerivationResult[DerivedExpr[To]] = {
-      DerivationResult.pure(
-        DerivedExpr.TotalExpr[To](
-          ctx.src.asInstanceOfExpr[To]
-        )
-      )
-    }
+    def apply[From, To](implicit ctx: TransformerContext[From, To]): DerivationResult[DerivedExpr[To]] =
+      DerivationResult.log(s"Matched ${Type.prettyPrint[From]} <:< ${Type.prettyPrint[To]} derivation rule") >>
```
krzemin, line 14: Instead of this, how about generic log before calling 'apply', that we're expanding rule of specific name?

chimney/src/main/scala/io/scalaland/chimney/dsl/FlagsDsl.scala
```diff
@@ -133,6 +133,22 @@ private[dsl] trait FlagsDsl[UpdateFlag[_ <: TransformerFlags], Flags <: Transfor
   def disableImplicitConflictResolution: UpdateFlag[Disable[ImplicitConflictResolution[?], Flags]] =
     disableFlag[ImplicitConflictResolution[?]]
 
+  /** Enable printing the logs from the derivation process.
+   *
+   * @see [[https://scalalandio.github.io/chimney/TODO]] for more details
+   * @since 0.8.0
+   */
+  def enableDerivationLog: UpdateFlag[Enable[DerivationLog, Flags]] =
```
krzemin, line 141: How about `enableMacrosLogging`/`disableMacrosLogging`?

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformSubtypesRuleModule.scala
MateuszKubuszok: Depends how we want to log recursive derivations and customization, IMHO ideally each nesting/recursion should result in ins own indentation in logs, so wherever we would put these logs it would be in some place that isn't accidentally skipped when we didn `Rule[From, To](someSubsetOfRules)`.

chimney/src/main/scala/io/scalaland/chimney/dsl/FlagsDsl.scala
MateuszKubuszok: Sounds reasonable

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformSubtypesRuleModule.scala
krzemin: I don't think we should directly call `apply` method of another rules. Even if you do, you would need to maintain nesting/indentation manually.

I believe we need a better/more explicit way to call logic from another rules.

MateuszKubuszok: Directly not. But we reserve ourselves the right to call something like `Rule(someSubsetOfRules)` somehow filter out some rules in recursive calls. And then it will be nested and annotated.

krzemin (approved):

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformSubtypesRuleModule.scala
krzemin: We concluded to remodel Rule interface that enables:
- explicit rules sequence swap in the future, if we ever need it
- continuing expansion, with the current rules sequence, following result of the complex logic (potentially recursive expansion)

Issue #324, opened 2023-07-07
Fix implementations for chimney-macro-commons for edge cases in Scala 3


MateuszKubuszok:
Scala 2 provides us with some utilities like:

* `tpe.finalResultType`
* `method.typeSignatureIn(tpe)`
* `tpe.substituteTypes(typeParams, typeArgs)`

In Scala 3 we don't have them and the first implementation of cross-platform utilities used by macros don't handle edge cases well:

```scala
case class Foo[A](a: A, rec: Option[Foo[A]])
// when we apply A =:= String we should have properties:
val a: String
val rec: Option[Foo[String]]
// but instead we get:
val a: String
val rec: Option[Foo[A]] // <- A =:= String hasn't been applied
```

```scala
sealed trait Foo[A, B]
case class Impl[C, D](c: C, d: D) extends Foo[D, C]
// when we apply A =:= String, B =:= Int
Foo[String , Int]
// should have a Impl subtype
Impl[Int, String]
```

Another issue is that Scala 2 macros provide is with:

```scala
methodSymbol.isPublic
```

which lets us check if some method or variable would be publicly available. Scala 3 encourages us to check flags:

```scala
!(sym.flags.is(Flags.Private) || sym.flags.is(Flags.PrivateLocal) || sym.flags.is(Flags.Protected))
```

but it doesn't help in situation like:

```scala
class Foo(val a: Int, b: String)
```

where list of symbols containing both `a` and `b` would NOT have any symbol with a private flag, which could lead to assumption that `b` is available, yet an attempt to use `b` would result in compilation error.

This needs to be fixed to unblock Scala 3 release. Code which would have to be modified lives in `chimney-macro-commons`, in `scala-3` folder.

MateuszKubuszok:
It seems that https://github.com/scalalandio/chimney/pull/336 fixed all of the issues we saw during compilation. @jchyb found some issues with symbol ownership with `-Xcheck-macros` and errors format seem to changed (at least in the context of munit, but it might be bug) - these need to be addressed eventually, but they aren't a blocker for preview/milestone release.

MateuszKubuszok:
Released in 0.8.0-RC1.

Issue #302, opened 2023-05-04
Document Chimney internals


MateuszKubuszok:
Once https://github.com/scalalandio/chimney/issues/276 and https://github.com/scalalandio/chimney/issues/301 are implemented, the mechanics behind Chimney derivation could be documented.

By documentation I mean:

- [ ] documenting the purpose, use cases and behavior of things like: `TransformerContext`, `DerivationResult` `DerivedExpr`, etc as ScalaDoc comments
- [ ] documenting workarounds and all not so straightforward solutions in the platform-specific code to explain _why_ it was done this way as inline comments
- [ ] creating `DESIGN.md` describing Chimney internals
   - [ ] how `.into[]...transform` creates an expression and `TypeClass.define.derive` creates a type class
   - [ ] which parts happen in compile time and which in runtime, how constants and functions are stored and extracted
   - [ ] general idea behind derivation-rule-based-architecture with some examples of the rules (and notice that the exact list of rules will evolve over the time and might be different for each Scala/platform so sourcecode should be the only source of through about them)
   - [ ] in the above, mention how recursive derivation works, how summoning works, that there are priorities to summoning in PartialTransformers etc
   - [ ] how Patchers biggyback on Transformers

This would serve both us, to write down our decisions and rationales behind them, as well as a help to future contributors.

krzemin:
> creating section in Chimney documentation about Chimney internals 

Suggest keeping it out of the user facing docs. Let's have another file like `DESIGN.md` for such internal stuff.

Issue #280, opened 2023-03-10
Add tests and then official support for Scala 3 macros handling Scala 2.13 code and vice-versa


MateuszKubuszok:
Once support for features currently supported in 2.13 is provided for Scala 3, add tests and later implementations verifying work with Scala 3 specific features:

- cross-compilations with 2.13
  - Scala 2.13 with `-Ytasty-reader` should:
    - derive code for case classes compiled with Scala 3
      - including code using default values
    - derive code for sealed traits compiled with Scala 3
    - derive code for enums derives compiled with Scala 3 (parameterless `case` has is `val` instead of `case object`) 
    - derive code for Scala 3 class with `@BeanProperty` (annotation changed behavior in Scala 3)
  - Scala 3 should:
    - derive code for case classes compiled with Scala 2.13
      - including code using default values
    - derive code for sealed traits compiled with Scala 2.13
    - derive code for Scala 2.13 class with `@BeanProperty` (annotation changed behavior in Scala 3)
- derive Scala 3 specific things:
  - opaque types with extension methods (Scala 3 macros see extension methods on list of methods!)
  - optionally deriving code for union types - provided that cases similar to https://github.com/lampepfl/dotty/issues/12668 will be solved and generated matches would be safe, without `the type test for Type cannot be checked at runtime` issue

MateuszKubuszok:
It seems that cross-compilation 2.13 and 3 despite being advertised as working ("sbt sandwich") still does NOT work (https://github.com/sbt/sbt-projectmatrix/issues/77)

MateuszKubuszok:
Removed from 0.8.0 milestone, as this has a low priority and I am unable to add these test until sbt-projectmatrix fix "sbt sandwich".

MateuszKubuszok:
At this point I am not sure if this is useful to anyone (nobody asked for it anyway), so am closing it.

MateuszKubuszok:
It became possible with https://github.com/sbt/sbt-projectmatrix/pull/97 (release https://github.com/sbt/sbt-projectmatrix/releases/tag/v0.10.1)

Issue #207, opened 2021-04-02
Locally defined implicit transformers are ignored


jonathan-ostrander:
Chimney version: 0.6.1 (Latest)

This issue is best demonstrated with the following example:

```scala
import io.scalaland.chimney.dsl._
import io.scalaland.chimney.Transformer

case class Foo(a: Int, b: Int)
case class Bar(c: Int = 0)

case class FooWrapper(foo: Foo)
case class BarWrapper(bar: Option[Bar])

// warning below that localT is never used
implicit val t: Transformer[FooWrapper, BarWrapper] = {
  implicit val localT: Transformer[Foo, Bar] = foo => Bar(foo.a + foo.b)
  Transformer.define[FooWrapper, BarWrapper].withFieldRenamed(_.foo, _.baz).buildTransformer
}
// warning: local val localT in value t is never used
//         implicit val localT: Transformer[Foo, Bar] = foo => Bar(foo.a + foo.b)
//                      ^
// error: No warnings can be incurred under -Xfatal-warnings.

// no warning because Bar has a default value in its default constructor
implicit val t: Transformer[FooWrapper, BarWrapper] =
  Transformer.define[FooWrapper, BarWrapper].withFieldRenamed(_.foo, _.baz).buildTransformer

FooWrapper(Foo(1, 2)).transformInto[BarWrapper]
// BarWrapper(Bar(0))

// redefine wrapper transformer with a Transformer[Foo, Bar] outside of a local scope
implicit val fooBarT: Transformer[Foo, Bar] = foo => Bar(foo.a + foo.b)
implicit val newT: Transformer[FooWrapper, BarWrapper] =
  Transformer.define[FooWrapper, BarWrapper].withFieldRenamed(_.foo, _.baz).buildTransformer

FooWrapper(Foo(1, 2)).transformInto[BarWrapper]
// BarWrapper(Bar(3))
```

There are obvious workarounds. `disableDefaultValues` can be used when building the wrapper transformer, but this is annoying when defining transformers with code generated by ScalaPB because the `unknownFields` argument must have a default value specified. Transformers can always be defined outside of a local scope, but sometimes a local scope is desired if multiple transformers require different transformation behavior for a given type in the same package.

MateuszKubuszok:
This issue can be solved adding a [flag to scalapb](https://scalapb.github.io/docs/customizations/#file-level-options):

```
  preserve_unknown_fields: false // true by default in Proto 3
```

[...]

PR #325, opened 2023-07-07
Rewrite of the code with initial Scala 3 implementation, fixes #201

### LINKED ISSUE #201: Scala 3 support
Are there any plans to support Scala 3 or do the changes in Scala 3 make this difficult?

MateuszKubuszok:
DONE:
 * rewritten whole derivation code from Scala 2-macro-implementation-tied, mostly untyped and lacking significant naming to more platform-agnostic and typed utilities with better names
 * provided platform-specific implementations for platform-agnostic interfaces in Scala 2 and in Scala 3
 * wired Scala 3 DSL to Scala 3 macros
 * removed old macros
 * removed deprecated features
   * `TransformerF`s
   * `enableUnsafeOption` flag
   * tests and benchmarks for the above
* split `Transformer` into `Transformer` and `Transformer.AutoDerived`, to avoid
  1. potentially expanding macros in implicit lookup, because it was catching not only user-provided implicits
  2. checking that this implicit was created by autoderivation macro (which introduces unnecessary boxing and ignores configs)
  3. discarding obtained implicit when it was autoderived
  4. effectively wasting CPU work
* sorted code inside `internal` package into `internal.runtime`, `internal.compiletime.dsl`, `internal.compiletime.derivation`, etc
* improved error logging in a few places
* added the ability to display structured logs from macro expansion (`.enableMacrosLogging`)

TODO:

- [x] #324 (the only actual blocker, ~requires help from VL, ETA unknown~ fixed in #336)
- [x] check if some issues would become low hanging fruits (optional, we can do it after the first Scala 3 release)
  * #110 - this should be fixed on new architecture but we need a test
  * #166 - since we split `Transformer` into `Transformer` and `Transformer.AutoDerived` we could also have 2 methods: one using autoderived and manually provided Transformers and one using only manually provided, and this task would be done
  * #175 - this should be fixed on new architecture but we need a test
  * #200 - this should be fixed on new architecture but we need a test
  * #204 - we might check what is the issue and fix it in chimney-macro-commons
  * #207 - this should be fixed on new architecture but we need a test
  * Protobufs oneof mapping could be improved - https://github.com/scalalandio/chimney/pull/331
- [x] task that could be done in this or the next release
  * #280 - can be done later
  * #302 
- [x] cleanup before release
  - [x] hide methods `__updateConfig` and similar from users - https://github.com/scalalandio/chimney/pull/334
  - [x] simplyfy macros(?) (optional, since only we look at them, we can modify the code any time) - we can do it later
  - [x] address failing tests on Scala 3 - some of them are(?) bugs in compiler and/or munit (workaround? report?) - disabled for now #345 
  - [x] fix symbol ownership on Scala 3 - it fails `-Xcheck-macros` OR Scala3.js - delegated to ticker #346 
  - [x] enable linters on Scala 3 (`-Xcheck-macros`, `-Xunused*`) - #337
- [x] provide documentation for changes
  - [x] **breaking change** removed deprecated features - #339
  - [x] **breaking change** split of `Transformer` into `Transformer` and `Transformer.AutoDerived`, reasons for doing so - #339
  - [x] **breaking change** fixing #200 - default value will not be used if the right field do exist in source value, but conversion for it couldn't be derived - #339
  - [x] partially fixing #110 and #175 by allowing usage of getters to point to setter fields (when pairing exists) - #339 (only need to be mentioned in release announcement)
  - [x] Sphinx docs for `.enableMacrosLogging` - #339
  - [x] Scaladocs for Scala 3 DSL - #338 
  - [x] Scaladocs for `.enableMacrosLogging` linking to Sphinx - #339 
- [x] check benchmarks for performance regressions in generated code
- [x] ~check if rewrite affected how long macros compile~ some feedbacks from users might be better than our own checks
- [ ] remove `scala-3` from `ci.yml` and `docs.yml`


README.md
```diff
@@ -161,22 +163,11 @@ curl -s https://raw.githubusercontent.com/scalalandio/chimney/master/try-chimney
...
-sbt makeSite
-```
-
-HTML Documentation should be generated at `target/sphinx/html/index.html`.
-
-Alternatively use Docker:
+For building documentation locally you can use Docker:
 
 ```bash
 docker run --rm -v "$PWD/docs:/docs" sphinxdoc/sphinx:3.2.1 bash -c "pip install sphinx-rtd-theme && make html"
```
krzemin, line 173: With `Makefile` being removed it won't work

.github/workflows/ci.yml
```diff
@@ -59,7 +59,7 @@ jobs:
           apps: sbt
 
       - name: Run benchmarks
-        run: sbt "++2.13.10 benchmarks/Jmh/run -rf json -rff $(pwd)/$(git describe --tags --always).json" # TODO LB scala version from matrix
+        run: sbt "++2.13.11 benchmarks/Jmh/run -rf json -rff $(pwd)/$(git describe --tags --always).json" # TODO LB scala version from matrix
```
krzemin, line 62: scala version seems to be redundant, as `benchmarks` is now managed by sbt-projectmatrix

README.md
krzemin: brought it back

.github/workflows/ci.yml
krzemin: fixed

chimney-macro-commons/src/main/scala-3/io/scalaland/chimney/internal/compiletime/ExprPromisesPlatform.scala
```diff
@@ -0,0 +1,144 @@
...
+    private def freshTermName[A: Type](usageHint: UsageHint): ExprPromiseName =
+      freshTermName(TypeRepr.of[A].show(using Printer.TypeReprShortCode).toLowerCase, usageHint)
+    private def freshTermName[A: Type](expr: Expr[?], usageHint: UsageHint): ExprPromiseName =
+      freshTermName[A](toFieldName(expr), usageHint)
+
+    // Undoes the encoding of freshTermName so that generated value would not contain $1, $2, ... and weird
+    // dot-replacement - this makes generated fresh names more readable as it prevents e.g. typename$macro$1$2$3
+    private def toFieldName[A](expr: Expr[A]): String =
+      if expr.asTerm.toString.contains("$u002E") then {
+        println(expr.asTerm.toString)
```
krzemin, line 72: Looks like unfinished debugging. What should be the final version?

MateuszKubuszok: It can be removed - I was checking which workarounds from Scala 2 are necessary and which could be removed and this checked for something that could be removed but I hadn't removed this println.

MateuszKubuszok: Removed

krzemin (approved):

PR #326, opened 2023-07-08
Test that Java Bean getters can be used in withFieldRenamed for renaming into setter

### LINKED ISSUE #110: Enable .withField*** when writing to Java bean
There's a limitation (https://scalalandio.github.io/chimney/#Limitations) that you can't provide values for missing fields when writing to Java bean.

The task is to find out whether we can make `.withFieldConst`, `.withFieldComputed` and `.withFieldRenamed` effective for Java beans too.

### LINKED ISSUE #175: withFieldComputed is ignored when enableBeanSetters
The example works with version 0.4.2 but it doesn't compile in 0.5.*

```

import scala.beans.BeanProperty

case class Entity(field1: String, field2: Option[String])

class AnotherEntity {
  @BeanProperty
  var field1: String = _
  @BeanProperty
  var field2: String = _
}

object Main extends App {
  val entity = Entity("1", Some("2"))

  val anotherEntity = entity.into[AnotherEntity]
    .withFieldComputed(_.field2, _.field2.orNull)
    .enableBeanSetters
    .transform
}
```

With 0.4.2 chimney the example compiles and properly copies the values.
With 0.5.3 I have an error
```
Chimney can't derive transformation from hello.Entity to hello.AnotherEntity

java.lang.String
  derivation from entity.field2: scala.Option to java.lang.String is not supported in Chimney!


hello.AnotherEntity
  field2: java.lang.String - can't derive transformation from field2: scala.Option in source type hello.Entity
```

The issue can be related to https://github.com/scalalandio/chimney/issues/110 but the field is not missing.

Here is a complete example that reproduces the issue https://github.com/melgenek/chimney-playground

MateuszKubuszok:
Added test which shows that the current architecture allow `withFieldRenamed` and other `withField*` methods to use Java Beans getter name as a way of selecting Java Bean setter that should be used in derivation e.g.

```scala
caseClass.into[JavaBean]
  .withFieldRenamed(_.name, _.getSurname) // will use setSurname(caseClass.name)
  .transform
```

It should showcase that most common cases of #110 and #175 are already handled (the only unsupported case would be if target had a setter without matching getter).

While writing these test I found and fixed a but occurring when withFieldRenamed was used toether with other withField and messed up their RuntimeDataStore index calculation

### DIFF
--- a/chimney/src/test/scala/io/scalaland/chimney/TotalTransformerJavaBeansSpec.scala
+++ b/chimney/src/test/scala/io/scalaland/chimney/TotalTransformerJavaBeansSpec.scala
@@ -44,14 +44,15 @@ class TotalTransformerJavaBeansSpec extends ChimneySpec {
     }
   }
 
-  group("""setting .withFieldRenamed(_.from, _.getTo)""") {
+  group("""settings .withField*(_.getTo, ...) and .withFieldRenamed(_.from, _.getTo)""") {
 
     test("transform case class to Java Bean, allowing using getters as a way to rename into matching setters") {
       val source = CaseClassWithFlagRenamed("test-id", "test-name", renamedFlag = true)
       val target = source
         .into[JavaBeanTarget]
+        .withFieldConst(_.getId, source.id)
+        .withFieldComputed(_.getName, _.name)
         .withFieldRenamed(_.renamedFlag, _.isFlag)
-        .enableBeanSetters
         .transform
 
       target.getId ==> source.id
--- a/chimney/src/test/scala/io/scalaland/chimney/TotalTransformerJavaBeansSpec.scala
+++ b/chimney/src/test/scala/io/scalaland/chimney/TotalTransformerJavaBeansSpec.scala
@@ -44,6 +44,22 @@ class TotalTransformerJavaBeansSpec extends ChimneySpec {
     }
   }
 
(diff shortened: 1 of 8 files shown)


chimney/src/test/scala/io/scalaland/chimney/TotalTransformerJavaBeansSpec.scala
```diff
@@ -44,6 +44,23 @@ class TotalTransformerJavaBeansSpec extends ChimneySpec {
     }
   }
 
+  group("""settings .withField*(_.getTo, ...) and .withFieldRenamed(_.from, _.getTo)""") {
+
+    test("transform case class to Java Bean, allowing using getters as a way to rename into matching setters") {
+      val source = CaseClassWithFlagRenamed("test-id", "test-name", renamedFlag = true)
+      val target = source
+        .into[JavaBeanTarget]
+        .withFieldConst(_.getId, source.id)
```
krzemin, line 53: How about adding a case when getter is not paired with setter with the same name? We should expect compilation to fail.

MateuszKubuszok: Added test, it required some change because code without it could successfully compile if ignored field override would make it compile.

PR #328, opened 2023-07-08
Add into*Safe methods which disable auto-derivation

MateuszKubuszok:
Since we already split derivation into automatic and configured, we might as well reflect it in the interface and fix some long overdue issue.

I didn't have a better name than `*Safe` but we might think of something better.


MateuszKubuszok:
Replaced with https://github.com/scalalandio/chimney/pull/357

PR #331, opened 2023-07-10
Better protobuf oneof support

MateuszKubuszok:
TODO:

- [x] extract WrapperClass from ValueClass so that not-only AnyVals could be extracted this way
- [x] use WrapperClass to introduce fallbacks in SealedHierarchyToSealedHierarchy which either unwrap From subtype or wrap To subtype to handle Protobug oneof-like cases
- [x] add tests


chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformSealedHierarchyToSealedHierarchyRuleModule.scala
```diff
@@ -84,14 +84,57 @@ private[compiletime] trait TransformSealedHierarchyToSealedHierarchyRuleModule {
...
+                                      toSubtype.Underlying
+                                    ](
+                                      wrapper.unwrap(fromSubtypeExpr)
+                                    ).map(_.map(toUpcast))
+                                )
+                              case _ => None
+                            }
+                          // We're constructing:
+                          // case fromSubtypeExpr: $fromSubtype => Subtype(${ derivedToSubtypeInner } // or ${ derivedResultToSubtypeInner }.map(Subtype)
+                          lazy val fromSubtypeToSubtypeUpwrapped = toSubtype.Underlying match {
```
krzemin, line 120:
```suggestion
                          lazy val fromSubtypeToSubtypeUnwrapped = toSubtype.Underlying match {
```

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformSealedHierarchyToSealedHierarchyRuleModule.scala
```diff
@@ -84,14 +84,57 @@ private[compiletime] trait TransformSealedHierarchyToSealedHierarchyRuleModule {
...
-                            .orElse(
-                              deriveRecursiveTransformationExpr[fromSubtype.Underlying, To](
-                                fromSubtypeExpr
+                          lazy val subtypeToSubtype =
+                            deriveRecursiveTransformationExpr[fromSubtype.Underlying, toSubtype.Underlying](
+                              fromSubtypeExpr
+                            ).map(_.map(toUpcast))
+                          // We're constructing:
+                          // case fromSubtypeExpr: $fromSubtype => ${ derivedToSubtype } // or ${ derivedResultToSubtype }
+                          lazy val subtypeToTarget =
```
krzemin, line 93:
```suggestion
                          lazy val subtypeToBaseType =
```

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformSealedHierarchyToSealedHierarchyRuleModule.scala
```diff
@@ -84,14 +84,57 @@ private[compiletime] trait TransformSealedHierarchyToSealedHierarchyRuleModule {
...
+                                  ).map(_.map(wrapper.wrap)).map(_.map(toUpcast))
                               )
-                            )
+                            case _ => None
+                          }
+
+                          subtypeToSubtype
+                            .orElse(subtypeToTarget)
+                            .orElseOpt(fromSubtypeUnwrappedToSubtype)
+                            .orElseOpt(fromSubtypeToSubtypeUpwrapped)
```
krzemin, line 137: How about `fromSubtypeUnwrappedToBaseType`? Looks like it would complete all the cases, but we would first need a motivating test.

krzemin (changes requested):
- address or remove all TODOs
- review assertions in `WrapperClassType`
- more tests, especially related to actual scalapb-generated protobuf classes

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformSealedHierarchyToSealedHierarchyRuleModule.scala
MateuszKubuszok: Done

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformSealedHierarchyToSealedHierarchyRuleModule.scala
MateuszKubuszok: For now I changed it to
```
                          fromSubtypeIntoToSubtype
                            .orElse(fromSubtypeIntoToType)
                            .orElseOpt(fromSubtypeUnwrappedIntoToSubtype)
                            .orElseOpt(fromSubtypeIntoToSubtypeUnwrapped)
```

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformSealedHierarchyToSealedHierarchyRuleModule.scala
MateuszKubuszok: Renamed to something more meaningful

MateuszKubuszok:
> * address or remove all TODOs
> 
>     * review assertions in `WrapperClassType`
> 
>     * more tests, especially related to actual scalapb-generated protobuf classes

Existing TODOs got resolves, with exception of a new TODO/idea for the future.

Protobufs classes are (again) generated from PB files, and I added tests for converting typical oneofs with our improved sealed rule.

I fixed WrapperClassType assertion.

chimney/src/test/scala/io/scalaland/chimney/PBTransformationSpec.scala
```diff
@@ -178,9 +178,37 @@ class PBTransformationSpec extends ChimneySpec {
 
   group("transformer sealed traits generated from oneof") {
 
-    test("CustomerStatus (oneof sealed_value)") {
+    object HandleEmptyAutomatically {
+      // TODO: consider creating some protobuf integration module in the future
```
MateuszKubuszok, line 182: Actually, we could do this already:
* rename `protos` to `chimney-protobufs`
* add this utility there to the `main` scope
* move these tests to `protos`
* change the order of dependency: `chimney-protos` depending on `chimney` and make it publishable

chimney-macro-commons/src/main/scala-3/io/scalaland/chimney/internal/compiletime/datatypes/ValueClassesPlatform.scala
```diff
@@ -7,44 +7,47 @@ trait ValueClassesPlatform extends ValueClasses { this: DefinitionsPlatform =>
...
 
-      val primaryConstructor: Symbol = Option(sym.primaryConstructor).filter(_.isClassConstructor).getOrElse {
-        assertionFailed(s"AnyVal ${Type.prettyPrint[A]} expected to have 1 public constructor")
+      (getterOpt, primaryConstructorOpt, argumentOpt) match {
+        case (Some(getter), Some(primaryConstructor), Some(argument)) if !Type[A].isPrimitive =>
+          val argumentT = paramsWithTypes(A, primaryConstructor)(argument.name).asType.asInstanceOf[Type[Any]]
+          val inner = returnTypeOf[Any](A.memberType(getter)).asExistential
+          import inner.Underlying as Inner
+          assert(
+            argumentT =:= Inner,
```
krzemin, line 31: Wondering if we should check this by assertion or with if-else. I'm worried it can be hacked by the user by providing specially prepared class, something like:

```scala
class Foo(private val b: Int) {
  def a: String = b.toString
}
```

MateuszKubuszok:
It gets problematic to rebase it with all the small fixes, if something needs to be addressed we can do in in a separate PR.

PR #330, opened 2023-07-10
Rework DSL

MateuszKubuszok:
DONE:

- [x] create utilities for updating `[Partial]TransfromInto`/`[Partial]TransformerDefinition` without `__methods`
- [x] express these utilities and tests in `ChimneyExprs` and `ChimneyTypes`
- [x] rewrite Scala 2 and Scala 3 DSL to use shared implementation
- [x] remove `__methods` from DSL classes

PR #334, opened 2023-07-18
Hide internal helpers from DSL API

MateuszKubuszok:
* removes `__updateRuntimeData`, `__refineConfig`, `__addOverride`, `__addInstance` from public API
 * since macros need access to some public interface there is new `WithRuntimeDataStore` in `internal.runtime` package
 * since types cannot be inferred this way, we need to provide all type parameters: `From`, `To`, `Cfg <: TransformerFrg` and `Flags <: TransformerFlags` for every whitebox macro now
 * usage of type constructors with manual type application got replaced with summoning with `weakTypeTag`, with some workaround for providing `WeakTypeTag` as `String` singleton type from selector expression
 * `TransformerInto` and `PartialTransformerInto` no longer use vampire macros
 * replaces https://github.com/scalalandio/chimney/pull/330


krzemin (approved):
LGTM 👍

PR #336, opened 2023-07-22
Fix genBCode crashes and paramsWithTypes failures

jchyb:
Previously the generated code would try to compare a value with the type of case object which caused crashes when generating backend code.  

It also turns out that constructor methods for case classes can include a type parameter, even if the type param of the case class was already set, so that was taken into account in the `paramsWithTypes` method. 

I also assumed-out the test that caused the compilation to fail - the issue lies in `scala.compiletime.testing.typeCheckErrors` in the compiler, and it is going to take a lot of time to minimize and fix that - likely not anytime soon, unfortunately.

After this PR, most, if not all of the non-implementation-is-missing errors concern types in the errors not rendering correctly and missing a prefix. I believe it may be caused either by munit, `scala.compiletime.testing.typeCheckErrors`, or both, because they change depending on the location of mis-rendered types (usually things behave as expected if removed from munit `group` or `test`). I tried looking for a quick fix, `typeRepr.typeSymbol.fullName` was the closest but it produces some unexpected characters which the compiler likes to internally use.


MateuszKubuszok:
@jchyb I merged what you manged to to for now, we can address the remaining issues in the followup PRs, thank you!

PR #343, opened 2023-08-01
Fix Scala.js 3 test compilation

MateuszKubuszok:
Thanks to @jchyb we found out that:
* `@static` annotation in Scala 3 causes our build to fail with
  ```
  [error] ## Exception when compiling 73 sources to /Users/dev/Workspaces/GitHub/chimney/chimney/target/js-3/classes
  [error] java.lang.AssertionError: assertion failed: Trying to access the this of another class:  tree.symbol = trait TransformerDefinitionCommons, class symbol = module class extensions$package$
  ```
  later on, we can minimize the example and report it to Scala.js and either then restore `@static` in our code, or use different implementation for JVM and Native and different for JS
* test compilations fails with [errors in `PBTransformationSpec`](
[error] Referring to non-existent class io.scalaland.chimney.examples.pb.order.PaymentStatus)
  - there must be somethings about ScalaPB and Scala 3, since it isn't the first problem I hade with it. I moved `PBTransformationSpec` tests to `protobufs` module and changed the direction of dependencies. This for some reason solved the issue and additionally let us remove `excludeDependencies` workaround
  - PB tests have to avoid Scala 3 syntax since we cannot selectively disable it and ScalaPB is not supporting them


jchyb:
Just checked, and the failing assertion causing the crash scalajs crash is similar to https://github.com/lampepfl/dotty/issues/17332, so a separate minimization should not be necessary

PR #353, opened 2023-08-29
Refactored SealedToSealed rule, re-enabled coproduct ambiguity check n Scala 3, fixed errors with too conservative checks

MateuszKubuszok:
* [x] split SealedToSealedRule body into smaller methods
* [x] fix overconservative check (it used to check that all From subtypes names are unique, and all To subtypes names are unique when only non-overriden-manually From types should have unambigous  To names matches)
* [x] re-enable check for that on Scala 3
* [ ] await fixing https://github.com/lampepfl/dotty/issues/18484


MateuszKubuszok:
Blocked by a bug in `scala.compiletime.testing.typeCheckErrors`, reproduced in https://gist.github.com/MateuszKubuszok/2723c2842b8bf969618f4bd751ba3041

MateuszKubuszok:
I need this refactors and fixes in other branches, so I cannot wait indefinite time for being able to have this one test being handled properly on Scala 3.

PR #354, opened 2023-08-29
Remove workarounds for Scala 3 error message checks in tests

MateuszKubuszok:
TODO:

 * [x] manually align the behavior of `Type.prettyPrint[T]` between Scala 2 and Scala 3
 * [x] fix methods generating `Transformer` and `PartialTransformer` type classes in Scala 3 - currently the pass `src` as the name of the expr, while they should use fresh name based on type
   * [x] ditto for Patcher 
 * [x] manually drop the suffix `$macro$1` from some freshterms on Scala 3 to align the behavior with Scala 2
   * [x] or better - change Scala 2 to add `$macro` to prefix so that both implementations would generate `$prefix$macro$n`, remove special cases in dropping `$macro$n` and add this suffix in tests 
 * [ ] await fixing https://github.com/lampepfl/dotty/issues/18484

### DIFF
--- a/chimney-macro-commons/src/main/scala-2/io/scalaland/chimney/internal/compiletime/ExprPromisesPlatform.scala
+++ b/chimney-macro-commons/src/main/scala-2/io/scalaland/chimney/internal/compiletime/ExprPromisesPlatform.scala
@@ -37,15 +37,16 @@ private[compiletime] trait ExprPromisesPlatform extends ExprPromises { this: Def
 
     private def freshTermName(prefix: String): ExprPromiseName =
       c.internal.reificationSupport.freshTermName(prefix.toLowerCase + "$")
+
     private def freshTermName(tpe: c.Type): ExprPromiseName =
       freshTermName(tpe.typeSymbol.name.decodedName.toString.toLowerCase)
     private def freshTermName(srcPrefixTree: Expr[?]): ExprPromiseName =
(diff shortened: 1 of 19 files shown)


MateuszKubuszok:
Blocked by a bug in scala.compiletime.testing.typeCheckErrors, reproduced in https://gist.github.com/MateuszKubuszok/055cef0b01a24fa430ff8aaf8a2caca0

MateuszKubuszok:
After some discussions and investigations, it seems that it's not a bug but an intended behavior. To fix all tests and have the same error messages between 2 and 3 I need to:

 * manually align the behavior of `Type.prettyPrint[T]` between Scala 2 and Scala 3
 * fix methods generating `Transformer` and `PartialTransformer` type classes in Scala 3 - currently the pass `src` as the name of the expr, while they should use fresh name based on type
 * manually drop the suffix `$macro$1` from some freshterms on Scala 3 to align the behavior with Scala 2

MateuszKubuszok:
Ultimately, showing `$macro$1` in outputs would make things harder to tests, maybe in the future we will remove it only from tests messages, but for now we remove it always from macro logs. Additionally, I re-ignored 1 test on Scala 3 which fails due to a compiler bug https://github.com/lampepfl/dotty/issues/18484 . Since these improvements will be helpful in other work needed for 0.8.0 I merge this and the last test will be un-ignored when fix will be available.

PR #391, opened 2023-09-26
[Not for 0.8.0] Create Java Collections module

MateuszKubuszok:
- [x] turn chimney-java-collections into publishable module
- [x] create instances for Java collections
  - [x] `java.util.Optional`
    - [x] test for total 
    - [x] test for partial
  - [x] `java.util.Iterable`
    - [x] test for total 
    - [x] test for partial
  - [x] `java.util.Enumeration`
    - [x] test for total 
    - [x] test for partial
  - [x] `java.util.Collection`
    - [x] test for total 
    - [x] test for partial
  - [x] `java.util.BitSet`
    - [x] test for total 
    - [x] test for partial
  - [x] `java.util.Dictionary`
    - [x] test for total 
    - [x] test for partial
  - [x] `java.util.Map`
    - [x] test for total 
    - [x] test for partial
  - [x] `java.util.Properties`
    - [x] test for total 
    - [x] test for partial
  - [x] `java.util.stream.Stream`
    - [x] test for total 
    - [x] test for partial
- [x] test that conversions produce correct error messages for partials
- [ ] create Java collections documentation

### DIFF
--- a/chimney-java-collections/src/main/scala/io/scalaland/chimney/javacollections/internal/TransformOrUpcast.scala
+++ b/chimney-java-collections/src/main/scala/io/scalaland/chimney/javacollections/internal/TransformOrUpcast.scala
@@ -1,0 +1,22 @@
+package io.scalaland.chimney.javacollections.internal
+
+import io.scalaland.chimney.Transformer
+
+/** Since [[io.scalaland.chimney.Transformer]] is NOT automatically provided for identity/subtype transformation,
+  * and we want to allow such things without enabling whole recursive auto-derivation we use this trick.
+  */
+trait TransformOrUpcast[From, To] {
+  def transform(src: From): To
+}
+object TransformOrUpcast extends TransformOrUpcastLowPriority {
+
+  implicit def userProvidedTransformerExists[From, To](implicit
+      transformer: Transformer[From, To]
+  ): TransformOrUpcast[From, To] =
+    from => transformer.transform(from)
+}
+private[internal] trait TransformOrUpcastLowPriority {
+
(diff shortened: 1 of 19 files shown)


MateuszKubuszok:
> For now however I am skipping on this effort as a lost a few hours and get nowhere with it.

Actually I managed to fix that! Refactored

chimney-java-collections/src/main/scala/io/scalaland/chimney/javacollections/internal/TransformOrUpcast.scala
```diff
@@ -0,0 +1,22 @@
+package io.scalaland.chimney.javacollections.internal
+
+import io.scalaland.chimney.Transformer
+
+/** Since [[io.scalaland.chimney.Transformer]] is NOT automatically provided for identity/subtype transformation,
+  * and we want to allow such things without enabling whole recursive auto-derivation we use this trick.
+  */
+trait TransformOrUpcast[From, To] {
```
MateuszKubuszok, line 8: Question: shouldn't such structure live in core? Would we reuse it somewhere else? If we want to enure binary compatibility we should decide where to put this utility before releasing it since shoving it around would break binary compatibility.

MateuszKubuszok:
I think that support for `java.util.Enum` can be done in a separate PR, as this one is already quite large and macro support is not needed for this PR to work.

MateuszKubuszok:
While it wasn't intended for 0.8.0 (more like 0.8.1) I decided to merge it, as it will be useful in convincing people to update.

Issue #400, opened 2023-10-16
Transforming into case class with mutable field breaks from 0.8.0


hughsimpson:
**Checklist**
- [x] I read the documentation at https://chimney.readthedocs.io/ and checked that the functionality exists
- [x] I verified that the behavior for my use case doesn't match the documentation
- [x] I checked the https://github.com/scalalandio/chimney/issues and haven't found the issue reported
- [x] I confirmed that the bug is not related to functionality that was deprecated: lifted transformers (`TransformerF`s) or `unsafeOption` flags

**Describe the bug**
Transforming into a class with a mutable field no longer works as of 0.8.0 - I did not find this in a list of removed functionality

**Reproduction**
The following works in 0.7.5, fails to compile on 0.8.0
```
object Main extends App {
  def fromRow(row: Bar): Foo = row.transformInto[Foo]
}

case class Foo(var string: String)

case class Bar(string: String)
```

**Expected behavior**
As per the documentation 'every class with a public primary constructor can be the target of the transformation.' - a case class with a mutable field is very much a class with a public primary constructor - I would expect this to continue to work as before

**Actual behavior**
```
~/chimney_min_bug/src/main/scala/chimney_min_bug/Main.scala:7:49: Chimney can't derive transformation from chimney_min_bug.Bar to chimney_min_bug.Foo
[error]
[error] chimney_min_bug.Foo
[error]   setStri(stri: java.lang.String) - no accessor named stri in source type chimney_min_bug.Bar
[error]
[error]
[error] Consult https://chimney.readthedocs.io for usage examples.
[error]
[error]   def fromRow(row: Bar): Foo = row.transformInto[Foo]
[error]                                                 ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
```

**Which Chimney version do you use**
Succeeds with 0.7.5, fails with 0.8.0

**Which platform do you use**
- [x] JVM
- [ ] Scala.js
- [ ] Scala Native

**If you checked JVM**
```
$ java -version
openjdk version "17.0.6" 2023-01-17
OpenJDK Runtime Environment GraalVM CE 22.3.1 (build 17.0.6+10-jvmci-22.3-b13)
OpenJDK 64-Bit Server VM GraalVM CE 22.3.1 (build 17.0.6+10-jvmci-22.3-b13, mixed mode, sharing)
```

**Additional context**
N/A

MateuszKubuszok:
It will be released in 0.8.1

PR #401, opened 2023-10-16
fix: transform into classes with var fields

hughsimpson:
After creating https://github.com/scalalandio/chimney/issues/400 it struck me that the fix was probably actually pretty simple


chimney/src/test/scala/io/scalaland/chimney/TotalTransformerMutableFieldSpec.scala
```diff
@@ -0,0 +1,100 @@
+package io.scalaland.chimney
+
+import scala.annotation.unused
+
+class TotalTransformerMutableFieldSpec extends ChimneySpec {
```
hughsimpson, line 5: This class is basically just a copy of `TotalTransformerDslsSeparationSpec` but with `var` fields on all Foo/Bar case classes. There may be a better place for the tests to go, but I'm not sufficiently familiar with the structure of the project... Nonetheless, these tests do pass without the accompanying fix

MateuszKubuszok: I'd say that for this issue adding a new entry in `IssuesSpec` would be enough. Just use enough transformations to cover all edge cases (transformInto/into.transform + maybe ditto for partials). If you want you can use either one test (`test("fix issue #400")`) or separate tests but grouped (`group("fix issue #400") { /* tests */ }`).

MateuszKubuszok:
Hey, great thanks for the reproduction and fix! I enabled the tests and all of them pass, so I would happyly merge it somewhere later this week. Congratulations on your first contribution!

chimney/src/test/scala/io/scalaland/chimney/TotalTransformerMutableFieldSpec.scala
hughsimpson: Ah great that sounds much neater, I'll rework the tests a little today then. Thanks for your quick response!

hughsimpson: Pushed neater tests 👍

MateuszKubuszok:
> @MateuszKubuszok pushed up some neater tests if you can retrigger the CI 🙏

Done!

In general LGTM, the only reason I am not merging this PR (and others) is that GH Actions are eager to break on our benchmarking environment and I would like to have a benchmark for 0.8.0 succeed to finish and get published before I merge new commits. But if that would take too long to fix I would skip it.

MateuszKubuszok (approved):

MateuszKubuszok:
> Oh yeah just looked. No idea what's going on there tbh, looks on the surface like something running out of memory for only 10kb but that obviously can't really be it unless there's a lot of those 10kb things

That's out own runner that we have to maintain ourselves. (For comparable benchmarks we have to have the same environment). It has some issues that we need to resolve.

MateuszKubuszok:
Merged, I am not sure when I'll release it, as there are a few more fixes I want to add to master, but it surely will be a part of 0.8.1. Thank you again for your contribution!

PR #444, opened 2023-12-13
Better support for collections outside Scala's std lib

MateuszKubuszok:
Java collections module introduced on 0.8.0 solved some problems:

 * provided support for Java optional type (`java.util.Optional`) similar to support for `scala.Option`
 * provided support for Java collections

It has however a few drawbacks:

 * it uses `TransformerOrUpcast` type to avoid issues with automatic vs semiautomatic derivation (prevents having to pick between 2 imports, one for automatically derived internal types and one for only semiautomatically derived)
 * it only handles Scala StdLib <=> Java Collections, or Java Collections <=> Java Collections
 * if we introduced support for e.g. Cata NonEmpty collection we could convert Cats' NonEmptyList to Scala's List and Scala's List to Java's List but not Cats' NonEmptyList to Java's list which might be counterintuitive
 * problem would got even worse if someone wanted to add support for more collections from some other library

To handle it, the best solution would be to:

 * create a typeclass for optional types
 * create a typeclass for iterating over the collection
 * create a typeclass for creating the collection
 * define implicits using these typeclasses in core module (allowing to select automatic or semiautomatic support)
 * make Java Collections module only a collections of instances for iterating and building collections
 * allow creating instances for NonEmpty* Cats collection in some sane manner

This would be a breaking change, so it should not be merged as a part of 0.8.x but it can become a part of future 1.0.0-RC line.

TODO:

 - [x] define new type classes
 - [ ] define new implicits (both automatic and semiautomatic) based on these type classes
   - [ ] test that these implicits work and do not create conflicts!
 - [ ] define instances in Java Collections module
   - [ ] test them
 - [ ] remove old solution
   - [ ] optionally make some "old" types deprecated aliases to the new types
 - [ ] document changes in docs
   - [ ] document how to provide support for custom Option (update existing section) or custom collection (create a new section)

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/integrations/IteratorOf.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/integrations/IteratorOf.scala
@@ -1,0 +1,8 @@
+package io.scalaland.chimney.integrations
+
+// TODO: add deprecated type alias in io.scalaland.chimney.javacollections
+trait IteratorOf[A, CC] {
+  def iterator(collection: CC): Iterator[A]
+
+  final def foreach(collection: CC)(f: A => Any): Unit = iterator(collection).foreach(f)
+}
--- a/chimney/src/main/scala/io/scalaland/chimney/integrations/OptionalOf.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/integrations/OptionalOf.scala
@@ -1,0 +1,27 @@
+package io.scalaland.chimney.integrations
+
+trait OptionalOf[F[_]] {
+  def someOf[A](a: A): F[A]
+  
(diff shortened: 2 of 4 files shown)


MateuszKubuszok:
I have a better idea now:

 * define new type classes (iteration, total builder, partial builder)
 * define something similar to `IterableOrArray` in macros logic which would fetch these type classes
 * handle code generation inside a macro

This collection-awareness would let us define something like `withFieldComputed(_.foo.eachIndex.bar.eachMapValue.baz, a => ...)` in the future.

MateuszKubuszok:
Closed in favor of https://github.com/scalalandio/chimney/pull/505

PR #448, opened 2023-12-25
Change how logos are being displayed

MateuszKubuszok:
Display logo in circle, centered on both README and landing page of docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
-# <img src="gfx/chimney-logo.svg" alt="Chimney logo" height="60" />  Chimney
+<p style="text-align: center"><img src="gfx/chimney-logo-circle-dark.svg" alt="Chimney logo" height="250px" /></p>
 
-[![chimney Scala version support](https://index.scala-lang.org/scalalandio/chimney/chimney/latest.svg)](https://index.scala-lang.org/scalalandio/chimney/chimney)
+# Chimney
+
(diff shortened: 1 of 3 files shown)


NoemiRozpara:
The logo seems to not be centered in `README`: 
image

Also I'd suggest putting the plain version everywhere on Github - as GH can be either in light or dark mode. The dark circle doesn't look too good on the dark background. The intention was to use the dark circle on the light background and the other way round. So I'd say either the orange one, or none 🙂 The circle may look good in avatars and tiny previews ✨

MateuszKubuszok:
> The logo seems to not be centered in `README`: image

Ups, I pressed merge before this comment appeared. Yes, it seems that [GH actively removes `style` tags](https://stackoverflow.com/questions/44831505/which-inline-html-styles-does-github-markdown-accept) :/

> Also I'd suggest putting the plain version everywhere on Github - as GH can be either in light or dark mode. The dark circle doesn't look too good on the dark background. The intention was to use the dark circle on the light background and the other way round. So I'd say either the orange one, or none 🙂 The circle may look good in avatars and tiny previews ✨

Actually, I tested it dark mode, I haven't used light mode in years. [Same with docs](https://chimney--448.org.readthedocs.build/en/448/).

MateuszKubuszok:
@NoemiRozpara  Should be better now.

NoemiRozpara:
Awesome! 🙌

PR #475, opened 2024-03-05
Test macros 2.13x3 cross compilation

MateuszKubuszok:
TODO:

 - [x] update build
   - [ ] test locally that artifacts are not broken
 - [x] define 2.13 and 3 test cases
   - [x] case classes
   - [x] case classes with `@BeanProperty`
   - [x] classes with default values in constructor
   - [x] enums with parameters in any value
   - [x] enums without parameters in any value
   - [x] sealed traits
 - [ ] tests 2.13 converting 3 classes and vice versa
   - [ ] case classes
   - [ ] @BeanProperties as source/target
   - [ ] sealed traits to/from enums
   - [ ] default values in target type
 - [ ] documentation - describe what is possible and what isn't


MateuszKubuszok:
Well, it seems even with workarounds "sbt sandwich" still ends up with a broken classpath. :/

MateuszKubuszok:
Without https://github.com/sbt/sbt/issues/7405 this approach would not work.

MateuszKubuszok:
Replaced by #647

Issue #480, opened 2024-03-14
Automatic Coproduct Rename from/to Protobuf GeneratedEnum


saeltz:
**Checklist**

- [x] I read the documentation at https://chimney.readthedocs.io/ and checked that the functionality doesn't exists
- [x] I checked the https://github.com/scalalandio/chimney/issues and haven't found the feature already requested reported
- [x] I confirmed that the request is not related to functionality that was deprecated: lifted transformers (`TransformerF`s) or `unsafeOption` flags
- [x] I believe that this is actual feature request, rather than [a question how to use existing features](https://github.com/scalalandio/chimney/discussions)

**Describe the desired behavior**

#409 proposes to allow renaming of coproducts. I think this is a good idea if one wants to transform from one enum to another with only the names being different.

I propose to allow for an automatic renaming if using Protobuf. If you follow the official Protobuf guidelines, enums are prefixed with the name of the enum and all uppercase, e.g.

```
enum Field {
  FIELD_UNSPECIFIED = 0;
  FIELD_CASE_ONE = 1;
  FIELD_CASE_TWO = 2;
}
```

In Scala (3), one wants to follow the best practices and use an enum with different names, e.g.

```
enum Field:
  case One, Two
```

This requires to define a custom transformer with one `.withCoproductInstance` per case. Introducing `.withCoproductRenamed` would simplify the code a bit but would still require a custom transformer with all cases.

I propose to have automatic renaming from Protobuf GeneratedEnum to Scala 3 enum with CamelCase. Possibly as an opt-in with the special protobuf import.

**Additional context**

I know this is heavily opinionated. I would love to get to know to solve this problem differently. If you're willing to have this feature, I'd be able to spend some time on it. I think it'd need to be added in `TransformSealedHierarchyToSealedHierarchyRuleModule.scala`.

MateuszKubuszok:
Hey 👋, great thanks for suggestion!

I agree that such feature would be quite useful, but I would hesitate to enable it by default - virtually every feature that we ever added and had it turned on by default ended up with a bug report from someone who expected the code to fail/require manual intervention. I don't think that all such cases make sense (if one ends up with more configs than code they are helping to replace, it kind of misses the point), but this one feels like something that could be opt-in.

I recently started developing a feature allowing to customize [how fields and subtypes are compared](https://github.com/scalalandio/chimney/pull/478). It would address quite similar request for fields (#89) but also add an ability to customize comparison of subtypes/enum values.

[...]

PR #489, opened 2024-04-01
#480 Automatic Coproduct Rename from/to Protobuf GeneratedEnum

saeltz:
I'm not entirely happy with the `TransformedNamesComparison` because it matches too broadly. I'm open for better suggestions.

I want to achieve a comparison for the following:

```
enum Field {
  FIELD_UNSPECIFIED = 0;
  FIELD_CASE_ONE = 1;
  FIELD_CASE_TWO = 2;
}
```

From/To

```
enum Field:
  case One, Two
```

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/dsl/TransformedNamesComparison.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/dsl/TransformedNamesComparison.scala
@@ -44,6 +44,16 @@ object TransformedNamesComparison {
     def namesMatch(fromName: String, toName: String): Boolean = fromName.equalsIgnoreCase(toName)
   }
 
+  case object GeneratedProtobufEnumEquality extends TransformedNamesComparison {
+
+    def namesMatch(fromName: String, toName: String): Boolean = {
+      val fromNameWithoutPrefix = fromName.replace("_", "").reverse.toLowerCase
+      val toNameWithoutPrefix = toName.replace("_", "").reverse.toLowerCase
+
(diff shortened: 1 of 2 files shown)


chimney/src/main/scala/io/scalaland/chimney/dsl/TransformedNamesComparison.scala
```diff
@@ -44,6 +44,16 @@ object TransformedNamesComparison {
     def namesMatch(fromName: String, toName: String): Boolean = fromName.equalsIgnoreCase(toName)
   }
 
+  case object GeneratedProtobufEnumEquality extends TransformedNamesComparison {
```
MateuszKubuszok, line 47: I would either name it `CamelCaseUnderscoreCaseEquality` (to e.g. use this with Java enums as well, maybe it should be a default for it? Maybe it should be named `JavaEnumAware`?) or keep it in `chimneyProtobuf` module. I wouldn't put some integration-specific conventions in the core package.

MateuszKubuszok (changes requested):
Thank you for the effort! I left a comment.

MateuszKubuszok:
Hey @saeltz 👋 , do you still want to work on this?

saeltz:
Hey @MateuszKubuszok, this currently matches too broadly. For this special use case I would probably rather go for something like https://github.com/moia-oss/teleproto/pull/326 by @lbialy.

MateuszKubuszok:
@saeltz - if you want to use Chimney in your macro-logic, with [1.1.0](https://github.com/scalalandio/chimney/releases/tag/1.1.0) it would be possible to just use Chimney macros directly. Please, open a discussion if you are interested in this.

PR #519, opened 2024-04-27
Add EnumAware subtype name comparison and make it default

MateuszKubuszok:


MateuszKubuszok:
MiMa is right, I should add it without making this default

PR #533, opened 2024-05-13
Attempt to fix Scala 3 enum in snippet.sc, instead reproduce an issue with TypeRepr.of[A].memberType(subtype)

MateuszKubuszok:
When Scala CLI `snippet.sc` is used macro generates assertion errors when Scala 3 enum is used:

 * recommended way of calculating the type is `TypeRepr.of[A].memberType(subtypeSymbol)`
 * the current way is based on `subtypeSymbol.typeRef` which cannot be used for types defined in
   ```scala
   class SomeClass {
     type InnerType
   }
   ```

   and `enum` and each `case` becomes a type defined in another class when we use `snippet.sc` (interestingly, the issue does not appear for `case object`s nor `case class`es)
* however this recommended solution also upcasts parametherless cases, so that children of
  ```scala
   enum Foo:
     case Bar(value: Int)
     case Baz
  ```
  are resolved to `List(Foo.Bar, Foo)` (which breaks macro in another way).

Once this is solved, all Scala 3 snippets in documentation with enums - which required:
```scala
@main def example: Unit = {
// actual code
}
```
can have the workaround removed.

See:

 * https://github.com/scala/scala3/issues/20349
 * https://github.com/scala/scala3/issues/19825

for the context (and why it's _not_ so easy to "just fix in in the library").

### DIFF
--- a/chimney-macro-commons/src/main/scala-3/io/scalaland/chimney/internal/compiletime/ExprPromisesPlatform.scala
+++ b/chimney-macro-commons/src/main/scala-3/io/scalaland/chimney/internal/compiletime/ExprPromisesPlatform.scala
@@ -19,7 +19,22 @@ private[compiletime] trait ExprPromisesPlatform extends ExprPromises { this: Def
         case NameGenerationStrategy.FromExpr(expr)  => freshTermName[From](expr, usageHint)
       }
 
-    protected def createRefToName[From: Type](name: ExprPromiseName): Expr[From] = Ref(name).asExprOf[From]
+    protected def createRefToName[From: Type](name: ExprPromiseName): Expr[From] = {
+      val From = TypeRepr.of[From]
+
+      val fixedRef = for {
+        sym <- Option(From.typeSymbol)
+        if sym.flags.is(Flags.Enum | Flags.StableRealizable) && !sym.flags.is(Flags.JavaStatic)
+        parentSymbol <- From.baseClasses.find(_.flags.is(Flags.Enum | Flags.Sealed))
+        parentType = From.baseType(parentSymbol)
+        fixedFrom = Type.platformSpecific.fromUntyped(parentType.memberType(sym)).as_??
+      } yield {
+        import fixedFrom.Underlying as FixedFrom
+        Ref(name).asExprOf[FixedFrom].asInstanceOf[Expr[From]]
+      }
+
+      fixedRef.getOrElse(Ref(name).asExprOf[From])
+    }
 
     def createLambda[From: Type, To: Type, B](
         fromName: ExprPromiseName,
@@ -126,7 +141,10 @@ private[compiletime] trait ExprPromisesPlatform extends ExprPromises { this: Def
         if sym.flags.is(Flags.Enum) && (sym.flags.is(Flags.JavaStatic) || sym.flags.is(Flags.StableRealizable)) then
           // Scala 3's enums' parameterless cases are vals with type erased, so w have to match them by value
           // case arg @ Enum.Value => ...
-          CaseDef(Bind(bindName, Ident(sym.termRef)), None, body)
+          val fixedSym =
+            if !sym.flags.is(Flags.JavaStatic) then TypeRepr.of[From].typeSymbol.children.find(_.name == sym.name).get
+            else sym
+          CaseDef(Bind(bindName, Ident(fixedSym.termRef)), None, body)
         else if sym.flags.is(Flags.Module) then
           // case objects are also matched by value but the tree for them is generated in a slightly different way
           // case arg @ Enum.Value => ...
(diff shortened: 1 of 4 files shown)


chimney-macro-commons/src/main/scala-3/io/scalaland/chimney/internal/compiletime/TypesPlatform.scala
```diff
@@ -69,24 +69,30 @@ private[compiletime] trait TypesPlatform extends Types { this: DefinitionsPlatfo
...
-            val typeParamReprs: List[TypeRepr] = typeParamSymbols.map(_.name).map(appliedTypeByParam)
-            fromUntyped[A](subtype.typeRef.appliedTo(typeParamReprs)).as_?<[A]
-          // subtype is monomorphic
-          case _ =>
-            fromUntyped[A](subtype.typeRef).as_?<[A]
-        }
+        fromUntyped[A] {
+          val isScala3EnumButNotJavaEnum =
+            subtype.flags.is(Flags.Enum | Flags.StableRealizable) && !subtype.flags.is(Flags.JavaStatic)
+          if isScala3EnumButNotJavaEnum then TypeRepr.of[A].memberType(subtype)
```
MateuszKubuszok, line 75: Recommended fix - unfortunately makes parameterless cases upcasted :(

chimney-macro-commons/src/main/scala-3/io/scalaland/chimney/internal/compiletime/ExprPromisesPlatform.scala
```diff
@@ -19,7 +19,22 @@ private[compiletime] trait ExprPromisesPlatform extends ExprPromises { this: Def
...
+      val fixedRef = for {
+        sym <- Option(From.typeSymbol)
+        if sym.flags.is(Flags.Enum | Flags.StableRealizable) && !sym.flags.is(Flags.JavaStatic)
+        parentSymbol <- From.baseClasses.find(_.flags.is(Flags.Enum | Flags.Sealed))
+        parentType = From.baseType(parentSymbol)
+        fixedFrom = Type.platformSpecific.fromUntyped(parentType.memberType(sym)).as_??
+      } yield {
+        import fixedFrom.Underlying as FixedFrom
+        Ref(name).asExprOf[FixedFrom].asInstanceOf[Expr[From]]
+      }
```
MateuszKubuszok, line 34: Naive attempt to apply `TypeRepr.of[A].memberType(subtype)` fix once we already matched types. Does not work.

chimney-macro-commons/src/main/scala-3/io/scalaland/chimney/internal/compiletime/ExprPromisesPlatform.scala
```diff
@@ -126,7 +141,10 @@ private[compiletime] trait ExprPromisesPlatform extends ExprPromises { this: Def
         if sym.flags.is(Flags.Enum) && (sym.flags.is(Flags.JavaStatic) || sym.flags.is(Flags.StableRealizable)) then
           // Scala 3's enums' parameterless cases are vals with type erased, so w have to match them by value
           // case arg @ Enum.Value => ...
-          CaseDef(Bind(bindName, Ident(sym.termRef)), None, body)
+          val fixedSym =
+            if !sym.flags.is(Flags.JavaStatic) then TypeRepr.of[From].typeSymbol.children.find(_.name == sym.name).get
+            else sym
+          CaseDef(Bind(bindName, Ident(fixedSym.termRef)), None, body)
```
MateuszKubuszok, line 147: Same as above.

chimney-macro-commons/src/main/scala-3/io/scalaland/chimney/internal/compiletime/ExprsPlatform.scala
```diff
@@ -160,11 +160,15 @@ private[compiletime] trait ExprsPlatform extends Exprs { this: DefinitionsPlatfo
...
-        .replaceAll("\\$\\d+", "")
+      try
+        expr.asTerm
+          .show(using Printer.TreeAnsiCode)
+          // remove $macro$n from freshterms to make it easier to test and read
+          .replaceAll("\\$macro", "")
+          .replaceAll("\\$\\d+", "")
+      catch {
+        case e: StackOverflowError => expr.toString
+      }
```
MateuszKubuszok, line 171: For some reason, when working with cases that this PR attempts to fix, I sometimes get StackOverflow from `expr.asTerm.show(using Printer.TreeAnsiCode)` :O

chimney/src/test/scala-3/io/scalaland/chimney/TotalTransformerEnumSpec.scala
```diff
@@ -15,6 +15,25 @@ class TotalTransformerEnumSpec extends ChimneySpec {
...
+      val result = {
+        // implicit val cfg = TransformerConfiguration.default.enableMacrosLogging
+        (Color1.Red: Color1).transformInto[Color2] ==> Color2.Red
+        (Color1.Green: Color1).transformInto[Color2] ==> Color2.Green
+        (Color1.Blue: Color1).transformInto[Color2] ==> Color2.Blue
+      }
+    }
+
+    new Snippet().result
+  }
```
MateuszKubuszok, line 35: Poorly named test to check if fix worked without `sbt publishLocal` -> `scala-cli scripts/test-snippets.scala` trip.

MateuszKubuszok:
I might consider some workaround where we'd fetch `m: scala.deriving.Mirror.SumOf[A]` and extract subtypes from `m.MirroredElemLabels`, but it has its own corner cases.

MateuszKubuszok:
Until it's fixed in upstream it makes no sense to work on this issue. When it will, then we might need to start from scratch anyway, so closing for now.

PR #543, opened 2024-05-24
Make sealed trait/enum matching order deterministic

reimai:
Classfiles of generated transformers are slightly different, the order of sealed subclasses is non-deterministic.
We're using bazel as a build tool and it's important for build to produce an exact same file for the same input, otherwise it's dependent targets are re-evaluated.

### DIFF
--- a/chimney-macro-commons/src/main/scala-2/io/scalaland/chimney/internal/compiletime/datatypes/SealedHierarchiesPlatform.scala
+++ b/chimney-macro-commons/src/main/scala-2/io/scalaland/chimney/internal/compiletime/datatypes/SealedHierarchiesPlatform.scala
@@ -36,7 +36,7 @@ trait SealedHierarchiesPlatform extends SealedHierarchies { this: DefinitionsPla
       // calling .distinct here as `knownDirectSubclasses` returns duplicates for multiply-inherited types
       extractRecursively(Type[A].tpe.typeSymbol.asType).distinct.map(typeSymbol =>
         subtypeName(typeSymbol) -> subtypeTypeOf[A](typeSymbol)
-      )
+      ).sortBy(_._1)
     }
 
     private def symbolsToEnum[A: Type](subtypes: List[(String, ?<[A])]): Enum[A] =
--- a/chimney-macro-commons/src/main/scala-2/io/scalaland/chimney/internal/compiletime/datatypes/SealedHierarchiesPlatform.scala
+++ b/chimney-macro-commons/src/main/scala-2/io/scalaland/chimney/internal/compiletime/datatypes/SealedHierarchiesPlatform.scala
@@ -29,14 +29,18 @@ trait SealedHierarchiesPlatform extends SealedHierarchies { this: DefinitionsPla
         .toList
 
     private def extractSealedSubtypes[A: Type]: List[(String, ?<[A])] = {
+
       def extractRecursively(t: TypeSymbol): List[TypeSymbol] =
         if (t.asClass.isSealed) t.asClass.knownDirectSubclasses.toList.map(_.asType).flatMap(extractRecursively)
         else List(t)
 
+      val order =
+        Ordering.fromLessThan[Position]((a, b) => a.line < b.line || (a.line == b.line && a.column < b.column))
+
       // calling .distinct here as `knownDirectSubclasses` returns duplicates for multiply-inherited types
-      extractRecursively(Type[A].tpe.typeSymbol.asType).distinct.map(typeSymbol =>
-        subtypeName(typeSymbol) -> subtypeTypeOf[A](typeSymbol)
-      ).sortBy(_._1)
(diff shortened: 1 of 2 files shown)


chimney-macro-commons/src/main/scala-2/io/scalaland/chimney/internal/compiletime/datatypes/SealedHierarchiesPlatform.scala
```diff
@@ -36,7 +36,7 @@ trait SealedHierarchiesPlatform extends SealedHierarchies { this: DefinitionsPla
       // calling .distinct here as `knownDirectSubclasses` returns duplicates for multiply-inherited types
       extractRecursively(Type[A].tpe.typeSymbol.asType).distinct.map(typeSymbol =>
         subtypeName(typeSymbol) -> subtypeTypeOf[A](typeSymbol)
-      )
+      ).sortBy(_._1)
```
MateuszKubuszok, line 39: While it would makes compilation deterministic most of the time, it might be at odds with how Scala 3 generates `.children`. So if we are improving that, I'd prefer to have **both** of the implementations use something like [file-position-based ordering](https://github.com/scalalandio/enumz/blob/master/modules/enumz/src/main/scala/io/scalaland/enumz/internal/SumTypeEnumMacros.scala#L13).

I said "most of the time", because we had cases in the past when someone compiled:

```scala
sealed trait Foo
object Foo {
  object Bar {
    case object Impl extends Foo
  }
  object Baz {
    case object Impl extends Foo
  }
}
```

which is a valid Scala, where only 2 `knownDirectSubclasses` have the same name. So ordering for such case would still be non-deterministic. Ordering by position in file has no such issues and matches what most people would expect.

MateuszKubuszok (changes requested):
Hello 👋 Thank you for a contribution! 🙏 

I left a comment with change request.

chimney-macro-commons/src/main/scala-2/io/scalaland/chimney/internal/compiletime/datatypes/SealedHierarchiesPlatform.scala
reimai: that's a great point, thank you!

MateuszKubuszok: Thank you! 🙏 Could you add it here as well: https://github.com/scalalandio/chimney/blob/master/chimney-macro-commons/src/main/scala-3/io/scalaland/chimney/internal/compiletime/datatypes/SealedHierarchiesPlatform.scala#L33 ?

Just to make sure both implementation follow the same contract.

Since `symbol.pos` is `Option[Position]` I think that we could assume `None < Some(_)` and then `startLine` comparison -> `startColumn` comparison to make both the matching behavior of both implementations obvious.

reimai: Oh, completely forgot about scala3, fixed that

chimney-macro-commons/src/main/scala-3/io/scalaland/chimney/internal/compiletime/datatypes/SealedHierarchiesPlatform.scala
```diff
@@ -29,10 +29,14 @@ trait SealedHierarchiesPlatform extends SealedHierarchies { this: DefinitionsPla
         else if sym.flags.is(Flags.Module) then List(sym.typeRef.typeSymbol.moduleClass)
         else List(sym)
 
+      val order = Ordering.Option(Ordering.fromLessThan[Position] { (a, b) =>
+        a.startLine < b.startLine || (a.startLine == b.startLine && a.startColumn < b.startColumn)
+      })
```
MateuszKubuszok, line 34: Crap... it's supposed to work but in `compileErrors` check it [throws an `AssertionError`](https://github.com/scalalandio/chimney/actions/runs/9252556722/job/25451767651?pr=543) - IMHO it's a bug and `pos` should not return `Some(position)` if in its internal representation `Span` is `NoSpam` (so there is _no_ `start`/`end`, not even 0).

I'll report it later, but for now I think we are forced to do something like:

```
_.pos.filter(pos => scala.util.Try(pos.start).isSuccess)
```
to work around this bug :/

MateuszKubuszok:
Thanks again for the fix! 🙏

Since we already updated Scala 2.13 stdlib to 2.13.14 (which slightly changes the contracts) I'd like the next release to be 1.1.0, and I would also like to merge Scala Native 0.5.1 support with it, which is currently blocked.

Once I'll be able to merge it, I'll publish your changes as part of 1.1.0.

MateuszKubuszok:
Released as 1.1.0

PR #121, opened 2019-11-01
Add path names to `NotSupportedDerivation` compile errors

htmldoug:
Summary
=======
Adds path names to compile errors.

Before
------
```
error: Chimney can't derive transformation from com.rallyhealth.chopshop.provider.model.PersonLocation to com.rallyhealth.chopshop.provider.es6.generated.EsPersonLocation

scala.Boolean
  derivation from scala.Option to scala.Boolean is not supported in Chimney!

java.lang.String
  derivation from scala.Option to java.lang.String is not supported in Chimney!
  derivation from scala.Int to java.lang.String is not supported in Chimney!

scala.collection.immutable.Set
  derivation from scala.Option to scala.collection.immutable.Set is not supported in Chimney!

scala.collection.Seq
  derivation from scala.Option to scala.collection.Seq is not supported in Chimney!

Consult https://scalalandio.github.io/chimney for usage examples.
```

After
-----
```
error: Chimney can't derive transformation from com.rallyhealth.chopshop.provider.model.PersonLocation to com.rallyhealth.chopshop.provider.es6.generated.EsPersonLocation

scala.Boolean
  derivation from personlocation.networkEnrollment.isCapitated: scala.Option to scala.Boolean is not supported in Chimney!

java.lang.String
  derivation from personlocation.networkEnrollment.effectiveInterval.start: scala.Option to java.lang.String is not supported in Chimney!
  derivation from personlocation.networkEnrollment.pcpId: scala.Option to java.lang.String is not supported in Chimney!
  derivation from personlocation.networkEnrollment.roles: scala.Int to java.lang.String is not supported in Chimney!
  derivation from personlocation.painManagementTier: scala.Option to java.lang.String is not supported in Chimney!
  derivation from personlocation.providerDigestId: scala.Option to java.lang.String is not supported in Chimney!

scala.collection.immutable.Set
  derivation from personlocation.networksByRole: scala.Option to scala.collection.immutable.Set is not supported in Chimney!

scala.collection.Seq
  derivation from personlocation.preferredPlans: scala.Option to scala.collection.Seq is not supported in Chimney!

Consult https://scalalandio.github.io/chimney for usage examples.
```

PR #564, opened 2024-06-30
Support derivation to non-sealed trait

jtjeferreira:
allow derivation to non-sealed trait, but require runtime that some runtime overrides exists to evaluate the rule

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformSealedHierarchyToSealedHierarchyRuleModule.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformSealedHierarchyToSealedHierarchyRuleModule.scala
@@ -16,11 +16,9 @@ private[compiletime] trait TransformSealedHierarchyToSealedHierarchyRuleModule {
       (Type[From], Type[To]) match {
         case (SealedHierarchy(Enum(fromElements)), SealedHierarchy(Enum(toElements))) =>
           mapEachSealedElementToAnotherSealedElement(fromElements, toElements)
-        case (SealedHierarchy(_), _) =>
-          DerivationResult.attemptNextRuleBecause(
-            s"Type ${Type.prettyPrint[From]} is a sealed/enum type but ${Type.prettyPrint[To]} is not"
-          )
-        case (_, SealedHierarchy(_)) =>
+        case (SealedHierarchy(Enum(fromElements)), _) if !ctx.config.areOverridesEmpty =>
+          mapEachSealedElementToAnotherSealedElement(fromElements, List.empty)
+        case (SealedHierarchy(Enum(fromElements)), _) =>
           DerivationResult.attemptNextRuleBecause(
             s"Type ${Type.prettyPrint[To]} is a sealed/enum type but ${Type.prettyPrint[From]} is not"
           )
--- a/chimney/src/test/scala/io/scalaland/chimney/IssuesSpec.scala
+++ b/chimney/src/test/scala/io/scalaland/chimney/IssuesSpec.scala
@@ -760,4 +760,14 @@ class IssuesSpec extends ChimneySpec {
     import Issue498.*
     (Foo.Sub1("test"): Foo).into[Bar].withFieldConst(_.b, 1).transform ==> Bar.Sub1("test", 1)
   }
+
+  test("support non-sealed trait destination") {
+    (colors1.Green: colors1.Color)
+      .into[String]
+      .withSealedSubtypeHandled[colors1.Red.type](_ => "red")
+      .withSealedSubtypeHandled[colors1.Green.type](_ => "green")
+      .withSealedSubtypeHandled[colors1.Blue.type](_ => "blue")
+      .transform ==> "green"
+
+  }
 }


chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformSealedHierarchyToSealedHierarchyRuleModule.scala
```diff
@@ -16,11 +16,9 @@ private[compiletime] trait TransformSealedHierarchyToSealedHierarchyRuleModule {
       (Type[From], Type[To]) match {
         case (SealedHierarchy(Enum(fromElements)), SealedHierarchy(Enum(toElements))) =>
           mapEachSealedElementToAnotherSealedElement(fromElements, toElements)
-        case (SealedHierarchy(_), _) =>
-          DerivationResult.attemptNextRuleBecause(
-            s"Type ${Type.prettyPrint[From]} is a sealed/enum type but ${Type.prettyPrint[To]} is not"
-          )
-        case (_, SealedHierarchy(_)) =>
+        case (SealedHierarchy(Enum(fromElements)), _) if !ctx.config.areOverridesEmpty =>
+          mapEachSealedElementToAnotherSealedElement(fromElements, List.empty)
```
jtjeferreira, line 20: my first approach was like this but without the `if !ctx.config.areOverridesEmpty`. Only 3 tests failed, mostly because error messages were different:
* io.scalaland.chimney.PartialTransformerStdLibTypesSpec
* io.scalaland.chimney.IssuesSpec
* io.scalaland.chimney.PartialTransformerIntegrationsSpec

For example the test `io.scalaland.chimney.IssuesSpec.fix issue #121` error message changed from
```
Chimney can't derive transformation from io.scalaland.chimney.IssuesSpec.Foo to io.scalaland.chimney.IssuesSpec.Bar

io.scalaland.chimney.IssuesSpec.Bar
  maybeString: scala.collection.immutable.Seq[java.lang.String] - can't derive transformation from maybeString: scala.Option[scala.collection.immutable.Set[java.lang.String]] in source type io.scalaland.chimney.IssuesSpec.Foo
  nested: io.scalaland.chimney.IssuesSpec.BarNested - can't derive transformation from nested: io.scalaland.chimney.IssuesSpec.FooNested in source type io.scalaland.chimney.IssuesSpec.Foo

scala.collection.immutable.Seq[java.lang.String]
  derivation from foo.maybeString: scala.Option[scala.collection.immutable.Set[java.lang.String]] to scala.collection.immutable.Seq[java.lang.String] is not supported in Chimney!

java.lang.String
  derivation from foo.nested.num: scala.Option[scala.Int] to java.lang.String is not supported in Chimney!

io.scalaland.chimney.IssuesSpec.BarNested
  num: java.lang.String - can't derive transformation from num: scala.Option[scala.Int] in source type io.scalaland.chimney.IssuesSpec.FooNested

Consult https://chimney.readthedocs.io for usage examples.
```

to 

```
Chimney can't derive transformation from io.scalaland.chimney.IssuesSpec.Foo to io.scalaland.chimney.IssuesSpec.Bar

java.lang.String
  derivation from none: scala.None to java.lang.String is not supported in Chimney!

scala.collection.immutable.Seq[java.lang.String]
  derivation from none: scala.None to scala.collection.immutable.Seq[java.lang.String] is not supported in Chimney!

io.scalaland.chimney.IssuesSpec.Bar
  maybeString: scala.collection.immutable.Seq[java.lang.String] - can't derive transformation from maybeString: scala.Option[scala.collection.immutable.Set[java.lang.String]] in source type io.scalaland.chimney.IssuesSpec.Foo
  nested: io.scalaland.chimney.IssuesSpec.BarNested - can't derive transformation from nested: io.scalaland.chimney.IssuesSpec.FooNested in source type io.scalaland.chimney.IssuesSpec.Foo

scala.collection.immutable.Seq[java.lang.String]
  can't transform coproduct instance scala.None to scala.collection.immutable.Seq[java.lang.String]
  can't transform coproduct instance scala.Some[scala.collection.immutable.Set[java.lang.String]] to scala.collection.immutable.Seq[java.lang.String]

java.lang.String
  derivation from some: scala.Some[scala.Int] to java.lang.String is not supported in Chimney!

java.lang.String
  can't transform coproduct instance scala.None to java.lang.String
  can't transform coproduct instance scala.Some[scala.Int] to java.lang.String

io.scalaland.chimney.IssuesSpec.BarNested
  num: java.lang.String - can't derive transformation from num: scala.Option[scala.Int] in source type io.scalaland.chimney.IssuesSpec.FooNested

scala.collection.immutable.Seq[java.lang.String]
  derivation from some: scala.Some[scala.collection.immutable.Set[java.lang.String]] to scala.collection.immutable.Seq[java.lang.String] is not supported in Chimney!

Consult https://chimney.readthedocs.io for usage examples.
```

which would make it worse...

Since when the target is not a sealed trait, the user needs to configure the transformation, adding the `!ctx.config.areOverridesEmpty` will only trigger this rule if some runtime overrides exists. WDYT?

chimney/src/test/scala/io/scalaland/chimney/IssuesSpec.scala
```diff
@@ -760,4 +760,14 @@ class IssuesSpec extends ChimneySpec {
     import Issue498.*
     (Foo.Sub1("test"): Foo).into[Bar].withFieldConst(_.b, 1).transform ==> Bar.Sub1("test", 1)
   }
+
+  test("support non-sealed trait destination") {
```
jtjeferreira, line 764: let me know if a better place for this test exists

chimney/src/test/scala/io/scalaland/chimney/IssuesSpec.scala
```diff
@@ -760,4 +760,14 @@ class IssuesSpec extends ChimneySpec {
...
+    (colors1.Green: colors1.Color)
+      .into[String]
+      .withSealedSubtypeHandled[colors1.Red.type](_ => "red")
+      .withSealedSubtypeHandled[colors1.Green.type](_ => "green")
+      .withSealedSubtypeHandled[colors1.Blue.type](_ => "blue")
+      .transform ==> "green"
```
jtjeferreira, line 770: Hi again @MateuszKubuszok,

While working on this test, I suddenly realised that this is not much different than

```scala
(colors1.Green: colors1.Color) match {
  case colors1.Red => "red"
  case colors1.Green => "green"
  case colors1.Blue => "blue"
}
```

or if I wanted to use Transformer, for the subtypes I can do that:

```scala
(colors1.Green: colors1.Color) match {
  case red:colors1.Red.type => red.into[Foo].transform
  case green:colors1.Green.type => green.into[Foo].transform
  case blue:colors1.Blue.Type => blue.into[Foo].transform
}
```

so I am happy to drop this PR, since we can achieve the same the same result and with the same effort, using regular pattern matching...

jtjeferreira: yeah, that would work. I will close this PR

Issue #496, opened 2024-04-10
Exclude macro generated code from WartRemover checks


baldram:
Hi there! Thanks for working on this amazing library! I have a suggestion for improvement.
It's probably just one line of code added to the macro code.
The link to an example solution in another project is at the very bottom.

**Checklist**

- [x] I read the documentation at https://chimney.readthedocs.io/ and checked that the functionality exists
- [x] I verified that the behavior for my use case doesn't match the documentation
- [x] I checked the https://github.com/scalalandio/chimney/issues and haven't found the issue reported
  - [x] I checked the https://github.com/scala/bug/issues and haven't found the issue reported there - relevant for Scala 2 compiler crashes, since it might be a bug in the compiler
  - [x] I checked the https://github.com/lampepfl/dotty/issues and haven't found the issue reported there - relevant for Scala 3 compiler crashes, since it might be a bug in the compiler 
- [x] I confirmed that the bug is not related to functionality that was deprecated: lifted transformers (`TransformerF`s) or `unsafeOption` flags
- [x] I believe that this is actual bug, rather than [a question how to use existing features](https://github.com/scalalandio/chimney/discussions)

**Describe the bug**

The goal would be to be able to use Chimney in projects where strict rules are active.

We've successfully used Chimney for years; however, after migrating from 0.6.2 to 0.8.5, it began causing Wartremover errors, leading to the need for linter exclusions. This introduces extra noise and prevents some code aspects from being linter-verified.
Practical example:

```scala
  @SuppressWarnings(
    Array( // `.buildTransformer` from Chimney causes wartremover warning
      "org.wartremover.warts.Equals",
      "org.wartremover.warts.Null",
      "org.wartremover.warts.Var",
      "org.wartremover.warts.SeqApply"
    )
  )
  private[schedule] val contractScheduleRequestTransformer: PartialTransformer[ContractScheduleRequestDto, ContractScheduleRequest] =
    PartialTransformer
      .define[ContractScheduleRequestDto, ContractScheduleRequest]
      .withFieldComputed( ... ) 
      ...
      .buildTransformer
```

**Reproduction**

Creating a Scala-CLI snippet is challenging because it doesn't support SBT plugins. However, this applies to any use of Chimney (after migrating to a newer version) where the transformer will be sufficiently complex.
But if you want me to attach an SBT project here, please let me know.

**Expected behavior**

The goal would be to be able to use Chimney in projects where strict rules are active. \
[...]

PR #572, opened 2024-07-14
Provide Scala-version-agnostic way of reading macro settings

MateuszKubuszok:
- [x] Scala-version-agnostic `-Xmacro-settings`
- [x] allow overriding every single `Transformer` (`Boolean`) flag
  - [x] `chimney.transformer.InheritedAccessors=boolean`
  - [x] `chimney.transformer.MethodAccessors=boolean`
  - [x] `chimney.transformer.DefaultValues=boolean`
  - [x] `chimney.transformer.BeanSetters=boolean`
  - [x] `chimney.transformer.BeanSettersIgnoreUnmatched=boolean`
  - [x] `chimney.transformer.NonUnitBeanSetters=boolean`
  - [x] `chimney.transformer.BeanGetters=boolean`
  - [x] `chimney.transformer.OptionDefaultsToNone=boolean`
  - [x] `chimney.transformer.PartialUnwrapsOption=boolean`
  - [x] `chimney.transformer.NonAnyValWrappers=boolean`
  - [x] `chimney.transformer.ImplicitConflictResolution=Total/Partial/false`
  - [x] `chimney.transformer.MacrosLogging=boolean`
- [x] allow overriding every single `Patcher` (`Boolean`) flag
  - [x] `chimney.patcher.IgnoreRedundantPatcherFields=boolean`
  - [x] `chimney.patcher.IgnoreNoneInPatch=boolean`
  - [x] `chimney.patcher.MacrosLogging=boolean`
- [x] allow providing comma-separated list of values to `@SuppressWarnings(Array(...))` and `@NoWarn(...)`
  - [x] `chimney.SuppressWarnings=false/list`
  - [x] `chimney.NoWarn=false/true/msg`
- [x] docs
  - [x] add a section about global flags override under scope-flags
  - [x] add a section about all possible settings and default values

### DIFF
--- a/chimney-macro-commons/src/main/scala-2/io/scalaland/chimney/internal/compiletime/DefinitionsPlatform.scala
+++ b/chimney-macro-commons/src/main/scala-2/io/scalaland/chimney/internal/compiletime/DefinitionsPlatform.scala
@@ -11,5 +11,8 @@ trait DefinitionsPlatform
 
   val c: blackbox.Context
 
-  protected val isScala212 = scala.util.Properties.versionNumberString < "2.13"
+  protected val XMacroSettings: List[String] = c.settings
+
+  /** Useful for distinction between 2.12 and 2.13, when necessary. */
+  protected val isScala212: Boolean = scala.util.Properties.versionNumberString < "2.13"
 }
--- a/chimney-macro-commons/src/main/scala-2/io/scalaland/chimney/internal/compiletime/ExprPromisesPlatform.scala
+++ b/chimney-macro-commons/src/main/scala-2/io/scalaland/chimney/internal/compiletime/ExprPromisesPlatform.scala
@@ -8,6 +8,13 @@ private[compiletime] trait ExprPromisesPlatform extends ExprPromises { this: Def
 
   protected object ExprPromise extends ExprPromiseModule {
 
+    object platformSpecific {
+
+      def freshTermName(prefix: String): TermName =
+        // Scala 3 generate prefix$macro$[n] while Scala 2 prefix[n] and we want to align the behavior
+        c.internal.reificationSupport.freshTermName(prefix.toLowerCase + "$macro$")
+    }
+
     // made public for ChimneyExprsPlatform: Transformer.lift and PartialTransformer.lift
     def provideFreshName[From: Type](
         nameGenerationStrategy: NameGenerationStrategy,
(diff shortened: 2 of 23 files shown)


MateuszKubuszok:
I need to test somehow whether this change will not break https://github.com/scalalandio/chimney/issues/496 / https://github.com/scalalandio/chimney/pull/497. Testing it is a bit of a PITA, but the is nothing in the ADTs' API (both on Scala2 and on Scala 3) that would make me certain I used them the right way.

MateuszKubuszok:
(`sbt publish-local-for-tests`)

Sooo:

```scala
//> using plugin org.wartremover:::wartremover:3.1.7
// Scala 2:
//> using scala 2.13.14
//> using options -Xfatal-warnings -P:wartremover:traverser:org.wartremover.warts.Unsafe -Wmacros:both -Xmacro-settings:chimney.SuppressWarnings=none -Vmacro
//> using dep io.scalaland::chimney:1.3.0-6-g42ce88f-SNAPSHOT

case class Foo(a: Option[String], b: Option[Int])
case class Bar(a: String, b: Int, c: Long)

println(
scala.compiletime.codeOf(
Foo(Some("test"), Some(1024))
  .intoPartial[Bar]
  .withFieldConst(_.c, 1024L)
  .transform
)
)
```

always pass (even though `@SuppressWarnings` is disabled, and we can check with `-Vmacro` that it gets properly set up to be modified/left intact/removed) while

```scala
//> using plugin org.wartremover:::wartremover:3.1.7
// Scala 3:
//> using scala 3.3.3
//> using options -Xfatal-warnings -P:wartremover:traverser:org.wartremover.warts.Unsafe -Xmacro-settings:chimney.SuppressWarnings=org.wartremover.warts.Var
//> using dep io.scalaland::chimney:1.3.0-6-g42ce88f-SNAPSHOT

case class Foo(a: Option[String], b: Option[Int])
case class Bar(a: String, b: Int, c: Long)

println(
Foo(Some("test"), Some(1024))
  .intoPartial[Bar]
  .withFieldConst(_.c, 1024L)
  .transform
)
)
```

always fails (an in the source code annotations are always absent :/)

MateuszKubuszok:
After testing with sbt... Scala 2 works as expected... while Scala 3 always pass, whether annotations are applied or not :/

MateuszKubuszok:
Sooo, for now let's continue, it seems that behavior is unchanged (verified with newer commit and Scala 3), if we notice some changes in wartemover and Scala 3 we'll return to the this.

PR #590, opened 2024-08-22
Add the `Result#fromCatchingNonFatal`

danicheg:
This PR adds one utility method which is similar to `cats.syntax.EitherObjectOps#catchNonFatal`.


MateuszKubuszok:
Hello @danicheg! Thanks for the PR! I cannot promise there will 1.5.0 release soon, but I will surely mention you in release notes when the time comes :)

PR #591, opened 2024-08-23
Add `contramap` and `map` ops to the `Transformer` and `PartialTransformer`

danicheg:
### Motivation
This PR adds four methods `Transformer#contramap`, `Transformer#map`, `PartialTransformer#contramap` and `PartialTransformer#map`, to make transformers more composable and reusing existing transformers more concise and pleasant.

### Examples
- `Transformer#contramap`:
```scala
val stringTransformer: Transformer[String, Int] = _.length

case class Id(id: String)

implicit val idTransformer: Transformer[Id, Int] =
  stringTransformer.contramap(_.id)
```
- `PartialTransformer#map`:
```scala
val stringTransformer: PartialTransformer[String, Int] =
  PartialTransformer.fromFunction(_.length)

case class Length(length: Int)

implicit val toLengthTransformer: PartialTransformer[String, Length] =
  stringTransformer.map(id => Length(id))
```

### Why not use existing capabilities?
- To emulate `Transformer#contramap` users might do
```scala
val stringTransformer: Transformer[String, Int] = _.length

case class Id(value: String)

implicit val idTransformer: Transformer[Id, Int] =
  id => stringTransformer.transform(id.value)
```

- To emulate `PartialTransformer#map` users might do
```scala
val stringTransformer: PartialTransformer[String, Int] =
  PartialTransformer.fromFunction(_.length)

case class Length(length: Int)

implicit val toLengthTransformer: PartialTransformer[String, Length] =
  PartialTransformer.apply(id => stringTransformer.transform(id).map(Length(_)))
```

Well, the main reason is to be concise and to promote reusing existing transformers (at the user's site), keeping the DRY principle at its max. These combinators are also widely known in functional communities, making their usage transparent and pleasant. OTOH, there is the `chimney-cats` module, which provides the same functionalities. However, using it would require many redundant allocations (implicit syntaxes come with their costs), which might not be ideal for GCs, especially since Chimney is often used in hot-paths of applications (based on my experience).

I'd appreciate any feedback on that matter!

### DIFF
--- a/chimney-cats/src/main/scala/io/scalaland/chimney/cats/CatsPartialTransformerImplicits.scala
+++ b/chimney-cats/src/main/scala/io/scalaland/chimney/cats/CatsPartialTransformerImplicits.scala
@@ -51,17 +51,16 @@ trait CatsPartialTransformerImplicits {
 
       override def tailRecM[A, B](a: A)(
           f: A => PartialTransformer[Source, Either[A, B]]
-      ): PartialTransformer[Source, B] =
-        (src, failFast) => {
-          def run(a1: A) = partial.Result.fromCatching(f(a1).transform(src, failFast)).flatMap(identity)
-          @scala.annotation.tailrec
-          def loop(a1: A): partial.Result[B] = run(a1) match {
-            case partial.Result.Value(Left(a2)) => loop(a2)
-            case partial.Result.Value(Right(b)) => partial.Result.Value(b)
-            case errors                         => errors.asInstanceOf[partial.Result[B]]
-          }
-          loop(a)
+      ): PartialTransformer[Source, B] = { (src, failFast) =>
+        def run(a1: A) = partial.Result.fromCatching(f(a1).transform(src, failFast)).flatMap(identity)
+        @scala.annotation.tailrec
+        def loop(a1: A): partial.Result[B] = run(a1) match {
+          case partial.Result.Value(Left(a2)) => loop(a2)
+          case partial.Result.Value(Right(b)) => partial.Result.Value(b)
+          case errors                         => errors.asInstanceOf[partial.Result[B]]
         }
+        loop(a)
+      }
 
       override def raiseError[A](e: partial.Result.Errors): PartialTransformer[Source, A] = (_, _) => e
 
--- a/chimney-cats/src/main/scala/io/scalaland/chimney/cats/CatsTotalTransformerImplicits.scala
+++ b/chimney-cats/src/main/scala/io/scalaland/chimney/cats/CatsTotalTransformerImplicits.scala
@@ -39,15 +39,14 @@ trait CatsTotalTransformerImplicits {
 
       override def tailRecM[A, B](a: A)(
           f: A => Transformer[Source, Either[A, B]]
-      ): Transformer[Source, B] =
-        src => {
-          @scala.annotation.tailrec
-          def loop(a1: A): B = f(a1).transform(src) match {
-            case Left(a2) => loop(a2)
-            case Right(b) => b
-          }
-          loop(a)
+      ): Transformer[Source, B] = { src =>
+        @scala.annotation.tailrec
+        def loop(a1: A): B = f(a1).transform(src) match {
+          case Left(a2) => loop(a2)
+          case Right(b) => b
         }
+        loop(a)
+      }
 
       override def coflatMap[A, B](
           fa: Transformer[Source, A]
@@ -60,6 +59,6 @@ trait CatsTotalTransformerImplicits {
   implicit final def catsContravariantForTransformer[Target]: Contravariant[Transformer[*, Target]] =
     new Contravariant[Transformer[*, Target]] {
       def contramap[A, B](fa: Transformer[A, Target])(f: B => A): Transformer[B, Target] =
-        b => fa.transform(f(b))
+        fa.contramap(f)
     }
 }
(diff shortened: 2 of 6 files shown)


chimney-cats/src/main/scala/io/scalaland/chimney/cats/CatsPartialTransformerImplicits.scala
```diff
@@ -51,17 +51,16 @@ trait CatsPartialTransformerImplicits {
...
-      ): PartialTransformer[Source, B] =
-        (src, failFast) => {
-          def run(a1: A) = partial.Result.fromCatching(f(a1).transform(src, failFast)).flatMap(identity)
-          @scala.annotation.tailrec
-          def loop(a1: A): partial.Result[B] = run(a1) match {
-            case partial.Result.Value(Left(a2)) => loop(a2)
-            case partial.Result.Value(Right(b)) => partial.Result.Value(b)
-            case errors                         => errors.asInstanceOf[partial.Result[B]]
-          }
-          loop(a)
```
danicheg, line 63: These changes arrived from Scalafmt, I haven't changed this specifically :shrug:

chimney-cats/src/main/scala/io/scalaland/chimney/cats/CatsTotalTransformerImplicits.scala
```diff
@@ -39,15 +39,14 @@ trait CatsTotalTransformerImplicits {
...
       override def tailRecM[A, B](a: A)(
           f: A => Transformer[Source, Either[A, B]]
-      ): Transformer[Source, B] =
-        src => {
-          @scala.annotation.tailrec
-          def loop(a1: A): B = f(a1).transform(src) match {
-            case Left(a2) => loop(a2)
-            case Right(b) => b
-          }
-          loop(a)
```
danicheg, line 49: \+ https://github.com/scalalandio/chimney/pull/591#discussion_r1728883084

chimney/src/main/scala/io/scalaland/chimney/PartialTransformer.scala
```diff
@@ -58,6 +58,33 @@ trait PartialTransformer[From, To] extends PartialTransformer.AutoDerived[From,
...
+    * }}}
+    *
+    * @param f
+    *   a pure function that maps a value of `To` to `A`
+    * @return
+    *   new [[io.scalaland.chimney.PartialTransformer PartialTransformer]]
+    *
+    * @since 1.5.0
+    */
+  final def map[A](f: To => A): PartialTransformer[From, A] = new PartialTransformer[From, A] {
```
MateuszKubuszok, line 81: If there is `map` in partial, there could also be `mapPartial[A](f: To => partial.Result[A]): PartialTransformer[From, A]`.

Since `Transformer` has `contramap` the `PartialTransformer` should also get one.

chimney/src/main/scala/io/scalaland/chimney/Transformer.scala
```diff
@@ -30,6 +30,28 @@ trait Transformer[From, To] extends Transformer.AutoDerived[From, To] {
...
+    * @param f
+    *   a pure function that maps a value of `A` to `From`
+    * @return
+    *   new [[io.scalaland.chimney.Transformer Transformer]]
+    *
+    * @since 1.5.0
+    */
+  final def contramap[A](f: A => From): Transformer[A, To] = new Transformer[A, To] {
+    override def transform(src: A): To = self.transform(f(src))
+  }
```
MateuszKubuszok, line 54: Since `PartialTransformer` has `map` the `Transformer` should also get one.

MateuszKubuszok (changes requested):
Hello, thanks for another PR!

I think we need a parity of features where possible, so both `Transformer` and `PartialTransformer` should have both `map` and `contramap` if we're adding them.

Additionally, we would prefer these operations to not live in the main typeclass but more in some dedicated trait:

```scala
// docs
trait Transformer[From, To] extends AutoDerived[From, To] {
  // docs
  def transform(src: From): To
}
object Transformer {

  trait AutoDerived[From, To] extends TransformerOps[From, To] {
    def transform(src: From): To
  }

  trait TransformerOps[From, To] {
    def transform(src: From): To

    def map[A](f: To => A): Transformer[From, A] = ...
    def contramap[A](f: A => From): Transformer[A, To] = ...
  }
}
```

so that they could be called and be present in intellisense but not distract from the main use case of `Transformer`s (and `PartialTransformer`s) when looking in the source code.

chimney/src/main/scala/io/scalaland/chimney/PartialTransformer.scala
danicheg: I thought about this once again and don't see any issue with adding `contramap` to `PartialTransformer` and `map` to `Transformer`, respectively.

What about `mapPartial`, it might be useful indeed, but don't you mind if we would add it separately?

MateuszKubuszok:
If you are still interested in working on this I just merged https://github.com/scalalandio/chimney/pull/753, which should make it more doable on [2.0.0-development](https://github.com/scalalandio/chimney/tree/2.0.0-development) branch (summoning work in expected way without an extra type per type-class, no issues with deciding which one should be extended etc). It would be released as `2.0.0-M2` once 2.13.17 is released, so there is no rush.

MateuszKubuszok:
Picked this up for the 2.0.0 line in #913, with you (@danicheg) as the commit author. Since this branch is a few hundred commits behind and predates the Hearth backend migration, I rebuilt the change fresh on `2.0.0-development` rather than rebasing, and completed it per the review here: both `Transformer` and `PartialTransformer` now get `map` + `contramap` (plus `mapPartial`/`contramapPartial` for the partial one), and `Iso`/`Codec` get invariant `imap*` combinators. The combinators live as plain `final def`s on the type classes (the dedicated `*Ops` trait is no longer needed now that the hierarchy is flattened). Added unit tests + a snippet-tested cookbook section. Thanks for the original work here \U0001F64F

PR #603, opened 2024-09-27
Fix ClassCastException for lowercase objects in Scala 3

### LINKED ISSUE #479: ClassCastException thrown on Scala 3 when transforming sealed traits to Protobuf GeneratedEnum
**Checklist**

- [x] I read the documentation at https://chimney.readthedocs.io/ and checked that the functionality exists
- [x] I verified that the behavior for my use case doesn't match the documentation
- [x] I checked the https://github.com/scalalandio/chimney/issues and haven't found the issue reported
  - [x] I checked the https://github.com/scala/bug/issues and haven't found the issue reported there - relevant for Scala 2 compiler crashes, since it might be a bug in the compiler
  - [x] I checked the https://github.com/lampepfl/dotty/issues and haven't found the issue reported there - relevant for Scala 3 compiler crashes, since it might be a bug in the compiler 
- [x] I confirmed that the bug is not related to functionality that was deprecated: lifted transformers (`TransformerF`s) or `unsafeOption` flags
- [x] I believe that this is actual bug, rather than [a question how to use existing features](https://github.com/scalalandio/chimney/discussions)

**Describe the bug**

A `ClassCastException` is thrown when trying to transform a `sealed trait` into a Protobuf `GeneratedEnum`.

**Reproduction**

Run [the test at this commit of this repository](https://github.com/moia-oss/teleproto/blob/f72e40a521f7f6a02b5728db340dc9d9f0913243/src/test/scala/io/moia/protos/teleproto/ProtocolBuffersRoundTripTest.scala).

This doesn't necessarily happen at every run but on most of the runs. Repeatedly `clean` before running the test may yield a different result.

**Expected behavior**

No exception.

**Actual behavior**

```
java.lang.ClassCastException: class io.moia.protos.teleproto.ProtocolBuffersRoundTripTest$Color$Yellow$ cannot be cast to class io.moia.protos.teleproto.ProtocolBuffersRoundTripTest$Color$pink$ (io.moia.protos.teleproto.ProtocolBuffersRoundTripTest$Color$Yellow$ and io.moia.protos.teleproto.ProtocolBuffersRoundTripTest$Color$pink$ are in unnamed module of loader sbt.internal.LayeredClassLoader @7dfaa86d)
[info]   at io.moia.protos.teleproto.ProtocolBuffersRoundTripTest$$anon$2.transform(ProtocolBuffersRoundTripTest.scala:20)
[info]   at io.moia.protos.teleproto.ProtocolBuffersRoundTripTest$$anon$2.transform(ProtocolBuffersRoundTripTest.scala:37)
[info]   at io.moia.protos.teleproto.ProtocolBuffersRoundTripTest$$anon$1.transform$$anonfun$1(ProtocolBuffersRoundTripTest.scala:46)
...
```

What's especially weird is that the first line of the stack trace is commented out.

**Which Chimney version do you use**

1.0.0-M1

**Which platform do you use**

- [x] JVM
- [ ] Scala.js
- [ ] Scala Native

**If you checked JVM**

OpenJDK 21.0.2

**Additional context**

We reproduced that on two MacBooks Pro M1.

If you go to [the next commit of the branch](https://github.com/moia-oss/teleproto/blob/f8b92dc00df8c578b91421f9617974a11df72d41/src/test/scala/io/moia/protos/teleproto/ProtocolBuffersRoundTripTest.scala) I fixed the exception by converting the `sealed trait` into a Scala 3 `enum`. I still thought it'd be helpful for you to take a look.

jchyb:
(presumably) fixes #479

I haven't seen anything being broken by this change locally, but let's see if CI agrees...

### DIFF
--- a/chimney-macro-commons/src/main/scala-3/io/scalaland/chimney/internal/compiletime/ExprPromisesPlatform.scala
+++ b/chimney-macro-commons/src/main/scala-3/io/scalaland/chimney/internal/compiletime/ExprPromisesPlatform.scala
@@ -140,10 +140,6 @@ private[compiletime] trait ExprPromisesPlatform extends ExprPromises { this: Def
         // Scala 3's enums' parameterless cases are vals with type erased, so w have to match them by value
         // case arg @ Enum.Value => ...
         CaseDef(Bind(bindName, Ident(sym.termRef)), None, body)
-        else if sym.flags.is(Flags.Module) then
-        // case objects are also matched by value but the tree for them is generated in a slightly different way
-        // case arg @ Enum.Value => ...
-        CaseDef(Bind(bindName, Ident(sym.companionModule.termRef)), None, body)
         else
           // case arg : Enum.Value => ...
           CaseDef(Bind(bindName, Typed(Wildcard(), TypeTree.of[SomeFrom])), None, body)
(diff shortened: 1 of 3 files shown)


MateuszKubuszok:
Thanks @jchyb! 🙏

Issue #248, opened 2023-02-11
Policy for checking which fields were used for transformation


MateuszKubuszok:
Some users would find it useful to be able to fail compilation if some/all fields from source were not used during derivation. Example of such thing was asked in this StackOverflow question: https://stackoverflow.com/questions/73927120/chimeny-scala-validate-source .

Such feature was implemented in [ducktape](https://github.com/arainko/ducktape/releases/tag/v0.1.3) so alternative libraries already decided that it is worth providing.

We could consider adding some/all of the setting:

* `.requireSourceFieldsUsed`
* `.requireSourceFieldsUsed(_.fieldA, _.fieldB, _.fieldC)`
* `.requireSourceFieldsUsedExcept(_.fieldA, _.fieldB)`
* `.requireTargetSubtypesUsed`
* `.requireTargetSubtypeUsed[SubType]`
* `.requireTargetSubtypeNotUsed[SubType]`

arainko:
Just to clarify, the thing implemented in `ducktape` is only applicable to the `Field.allMatching` config option, where it'll fail to compile if you pass in a case class that matches none of the destination fields (but only for the case class passed to this config option) like this:

```scala
test("Field.allMatching reports a compiletime failure when none of the fields match") {
    final case class Source(int: Int, str: String, list: List[String])
    final case class FieldSource(int: Long, str: CharSequence, list: Vector[String])

    assertFailsToCompileWith {
      """
      val source = Source(1, "str", List("list-str"))
      val fieldSource = FieldSource(1L, "char-seq", Vector("vector-str"))
      source.into[Source].transform(Field.allMatching(fieldSource))
      """
    }("None of the fields from FieldSource match any of the fields from Source.")
  }
```
But it doesn't affect other parts of derivation, so you'd be the trailblazers here ❤️

MateuszKubuszok:
We could 😉 But it would be seriously helpful to know more about use cases of this feature. I saw some options named "policy" in [Java's MapStruct](https://mapstruct.org/documentation/stable/reference/html/) but I have no idea what users could find useful, so any feedback about this is welcome!

For now we are mostly adding bugs and ideas to backlog to keep tabs on them, since we have a few things already planned (https://github.com/scalalandio/chimney/issues/239) and they will take some time 😄

SevDan:
[...]

Issue #601, opened 2024-09-25
New flag `ExhaustSource`


nimatrueway:
**Checklist**

- [x] I read the documentation at https://chimney.readthedocs.io/ and checked that the functionality doesn't exist
- [x] I checked the https://github.com/scalalandio/chimney/issues and haven't found the feature already requested reported
- [x] I confirmed that the request is not related to functionality that was deprecated: lifted transformers (`TransformerF`s) or `unsafeOption` flags
- [x] I believe that this is actual feature request, rather than [a question how to use existing features](https://github.com/scalalandio/chimney/discussions)

**Describe the desired behavior**

```scala
case class RequestModel(name: String, age: Int)
case class DatabaseModel(name: String)

val toDatabaseModelT =
  Transformer
    .define[RequestModel, DatabaseModel]
    .enableExhaustSource // proposed flag, results in compile error: `age` is not used
    .withSourceFieldIgnored(_.age) // proposed new utility method, fixes the compile error
    .buildTransformer
}
```

Thanks for the great library!

I'd propose that we add a new flag `ExhaustSource` to the transformer that emits compiler error if not all source fields were mapped to a field in the destination model. This is useful when you'd like to enforce a 1:1 mapping between source and destination field.

**Use case example**

As the `From` model evolves you'd like to rely on the compiler to tell you what new fields need too be added to the `To` model

- When `From` is API request model and `To` is database persistence model
- Same but with database model as `From` and response model as `To`

**How it relates to existing features**

Preserves the existing behaviour. The new feature just adds an stricter check if enabled.

**Additional context**

As a workaround in our company we build another transformer but in the opposite direction and never use it, it is 
added only to emit compile error if an extra field of `From` is not used. This has led to significant redundant code which would not be needed if this feature existed.

If the maintainer sanity checks this idea, I might give it a shot myself as a hobby project.

MateuszKubuszok:
I'm afraid it might be a bit more complex than just boolean flag since just OTOH the valid behavior could be:

  - ignore unused source fields, 
  - fail on any unused source fields in top-level source value, ignore unused in nested transformations
  - explicitly allow a list of unused source values, require all the others to be used
  - explicitly allows a list of fields to use, require all the others unused

and for subtypes (because they can only be unused on the target side)

  - ignore unused enum subtypes
  - fail on any unused enum subtype
  - ...

[...]

Issue #115, opened 2019-07-11
Support for mapping multiple case classes


nsutcliffe:
Presuming I had the following:

```scala
case class Container1(a: Int)
case class Container2(b: String)
case class MergedContainers(a: Int, b: String)
```

It would be nice to do something like either of the following:
```scala
(Container1(1),Container2("HI")).transformInto[MergedContainers]
```

Or perhaps:
```scala
case class NestedContainers(container1: Container1, container2: Container2)
NestedContainers(Container1(1),Container2("HI")).transformInto[MergedContainers]
```

krzemin:
Having case class that combines containers (`NestedContainer`), this should be also achievable by first proposal: https://github.com/scalalandio/chimney/issues/100#issuecomment-480563946

ittaiz:
Do you think this will be implemented sometime soon? Had a need for it a few times and used hacks but I have a feeling I’ll have a blocker in the next month or two and this could really come in handy

krzemin:
@ittaiz many people ask for this feature recently - I think it should be next top priority. But it requires a lot of time to design it correctly, so I can't guarantee any fixed time frame.

ittaiz:
Sure thing, I complete understand.
I appreciate the intent and thank you for your work!

michaelbilow:
Ah, yeah, this is the feature I've been looking for. I imagine it's on the hard side to implement. Many thanks for your work on chimney @MateuszKubuszok

MateuszKubuszok:
I have _some_ idea how to do it, but it will be a significant amount of work.

I was thinking about prioritizing 1.0.0-RC first, to not delay it anymore, and then( later in 2024) starting to work on such merging transformation (it would allow us to internally reimplement `Patcher`s as such transformer). The challenges for this feature are that:

 * it would have to merge values _recursively_ because that's where lies its true usefulness
 * it would have to allows merging and transforming an arbitrary number of values
 * it would have to cooperate with all the existing overrides (`.withField*`, `.enable*`)
 * it would have to preserve the semantics of existing code, and make sure that existing tests will still pass

I have some an idea how to do it, but it would require refactoring a few sensitive pieces of code, writing a bit more pieces of a new code, and creating a tons of tests to avoid [feature interaction bugs](https://www.hillelwayne.com/post/feature-interaction/), because at this point it's really easy to write code which might look okayish, pass a simple test, compile (obviously) but provide an unexpected behavior when combined with some other feature.

[...]

PR #608, opened 2024-09-29
implement requireSourceFieldsUsedExcept

nimatrueway:
An implementation of `requireSourceFieldsUsedExcept(_.field1, _.field2..)` as described in https://github.com/scalalandio/chimney/issues/248

TODO/Future Work:
- [x] move usedSourceFields out from DerivationResult state and move it to `ProductToProduct` class
- [ ] implement a vararg version of `def ignoreUnusedField(selector, selectors*)` as Mateus suggested
- [ ] add `FailOnSpecifiedUnused` subclass for `ActionOnUnused`
  - [ ] implement a new setting to DSL `def requireUsedField(selector, selectors*)` as Mateus suggested
  - [ ] add `RequireUsedField` to `TransformerOverrides` to add to list of fields to be checked if `FailOnSpecifiedUnused` is enabled

### DIFF
--- a/chimney/src/main/scala-2/io/scalaland/chimney/internal/compiletime/ChimneyTypesPlatform.scala
+++ b/chimney/src/main/scala-2/io/scalaland/chimney/internal/compiletime/ChimneyTypesPlatform.scala
@@ -96,6 +96,9 @@ private[compiletime] trait ChimneyTypesPlatform extends ChimneyTypes { this: Chi
     val PreferPartialTransformer: Type[io.scalaland.chimney.dsl.PreferPartialTransformer.type] =
       weakTypeTag[io.scalaland.chimney.dsl.PreferPartialTransformer.type]
 
+    val FailOnUnused: Type[io.scalaland.chimney.dsl.FailOnUnused.type] =
+      weakTypeTag[io.scalaland.chimney.dsl.FailOnUnused.type]
+
     val RuntimeDataStore: Type[dsls.TransformerDefinitionCommons.RuntimeDataStore] =
       weakTypeTag[dsls.TransformerDefinitionCommons.RuntimeDataStore]
 
@@ -249,15 +252,15 @@ private[compiletime] trait ChimneyTypesPlatform extends ChimneyTypes { this: Chi
             )
           }
       }
-      object RequireSourceFieldsExcept extends RequireSourceFieldsExceptModule {
+      object IgnoreUnusedField extends IgnoreUnusedFieldModule {
         def apply[
-            FromPathList <: runtime.PathList: Type,
-            Tail <: runtime.TransformerOverrides: Type
-        ]: Type[runtime.TransformerOverrides.RequireSourceFieldsExcept[FromPathList, Tail]] =
-          weakTypeTag[runtime.TransformerOverrides.RequireSourceFieldsExcept[FromPathList, Tail]]
-        def unapply[A](A: Type[A]): Option[(?<[runtime.PathList], ?<[runtime.TransformerOverrides])] =
-          A.asCtor[runtime.TransformerOverrides.RequireSourceFieldsExcept[?, ?]].map { A0 =>
-            (A0.param_<[runtime.PathList](0), A0.param_<[runtime.TransformerOverrides](1))
+          FromPath <: runtime.Path: Type,
+          Tail <: runtime.TransformerOverrides: Type
+        ]: Type[runtime.TransformerOverrides.IgnoreUnusedField[FromPath, Tail]] =
+          weakTypeTag[runtime.TransformerOverrides.IgnoreUnusedField[FromPath, Tail]]
+        def unapply[A](A: Type[A]): Option[(?<[runtime.Path], ?<[runtime.TransformerOverrides])] =
+          A.asCtor[runtime.TransformerOverrides.IgnoreUnusedField[?, ?]].map { A0 =>
+            (A0.param_<[runtime.Path](0), A0.param_<[runtime.TransformerOverrides](1))
           }
       }
     }
@@ -318,6 +321,15 @@ private[compiletime] trait ChimneyTypesPlatform extends ChimneyTypes { this: Chi
           def unapply[A](A: Type[A]): Option[?<[dsls.ImplicitTransformerPreference]] =
             A.asCtor[runtime.TransformerFlags.ImplicitConflictResolution[?]].map { A0 =>
               A0.param_<[dsls.ImplicitTransformerPreference](0)
+            }
+        }
+        object UnusedFieldPolicy extends UnusedFieldPolicyModule {
+          def apply[R <: dsls.ActionOnUnused: Type]
+          : Type[runtime.TransformerFlags.UnusedFieldPolicy[R]] =
+            weakTypeTag[runtime.TransformerFlags.UnusedFieldPolicy[R]]
+          def unapply[A](A: Type[A]): Option[?<[dsls.ActionOnUnused]] =
+            A.asCtor[runtime.TransformerFlags.UnusedFieldPolicy[?]].map { A0 =>
+              A0.param_<[dsls.ActionOnUnused](0)
             }
         }
         object FieldNameComparison extends FieldNameComparisonModule {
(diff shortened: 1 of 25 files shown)


chimney/src/main/scala-2/io/scalaland/chimney/internal/compiletime/ChimneyTypesPlatform.scala
```diff
@@ -406,6 +417,17 @@ private[compiletime] trait ChimneyTypesPlatform extends ChimneyTypes { this: Chi
...
+    object PathList extends PathListModule {
+      val Empty: Type[runtime.PathList.Empty] = weakTypeTag[runtime.PathList.Empty]
+      object List extends ListModule {
+        def apply[Head <: runtime.Path: Type, Tail <: runtime.PathList: Type]: Type[runtime.PathList.List[Head, Tail]] =
+          weakTypeTag[runtime.PathList.List[Head, Tail]]
+        def unapply[A](A: Type[A]): Option[(?<[runtime.Path], ?<[runtime.PathList])] =
+          A.asCtor[runtime.PathList.List[?, ?]]
+            .map(A0 => A0.param_<[runtime.Path](0) -> A0.param_<[runtime.PathList](1))
+      }
+    }
```
MateuszKubuszok, line 429: A list of paths would not be needed if instead this would apply a config, as many times as many paths there is in vararg. This would make less code to maintain in DSL and macros.

MateuszKubuszok (changes requested):
Hello, thank you for your PR!

I think it's really nice first iteration, showing that you really looked around the codebase 😄 

In general, for the sake of backward compatibility and limiting the amount of new types, and flexibility I would like to suggest several things:

 * separation of the general policy (`IgnoreUnusedFields`/`FailOnUnusedFields`/`FailOnUnusedFieldStrict`/etc) from its exceptions (`IgnoreField[Path, Cfg]`, `FailOnUnusedField[Path, Cfg]`) both in DSL, type-level config and config within macros
 * that would allow changing the policy globally or in the whole scope
 * allowed/rejected fields could be appended one by one, matching current conventions for other customizations
   * if needed there could be 3 possible overrides: ignore field, reject unused field, remove override for field

So that would mean:

 * storing field policy as a `TransformerFlag` similar to `ImplicitTransformerPreference` (allowing us to add more policiies in the future)
   * e.g. `.unusedFieldPolicy(FailOnUnused)`
 * adding exceptions to list... maybe requiring non-empty list of overrides
   * e.g. `def ignoreUnusedField(selector, selectors*)`
 * things would be stored in a slightly different way internally, mostly using `runtimeOverrides: Vector[(Path, TransformerOverride)]` which is already managing propagation of configs based on paths

Such DSL would require a but more keystrokes but it would better fit Chimney's existing conventions.

nimatrueway:
> Hello, thank you for your PR!
> 
> I think it's really nice first iteration, showing that you really looked around the codebase 😄
> ...

My pleasure! Thank you so much for the quick review and good ideas. I initially had a similar thought in https://github.com/scalalandio/chimney/issues/601, it's nice to hear it wasn't too far off.

I'll start developing the API change 🤓🚧

nimatrueway:
Hey @MateuszKubuszok, I addressed the architectural change you proposed (and almost all your comments). Namely I split the API into a flag and a setting to override configuration per field.

```scala
// implementation:done -> tracks all fields and forces compile error if any left unused
.enableUnusedFieldPolicy(FailOnUnused)

// implementation:done -> untracks a single field
// TODO: how do I implement the vararg version without introducing PathList type? could use your direction here
.withIgnoreUnusedField(_.b)
```

To avoid a API change in the future and preserve backward compatibility, it's probably best to fix the API in this PR. So I tried to formulate your vision. Here's the domain terms and phrases I imagined makes sense for this feature. Please share if you have any suggestion.

- I would like to `track` certain fields and `fail` transformation derivation if any are `unused`.
- I would like to `track` certain fields and spit out `warning` during transformation derivation if any are `unused`.
- I would like to `track` all fields of the source type by default.
- I would like to `track` none of fields of the source type by default.
- I would like to `ignore` certain fields if `unused`.
- I would like to `track` certain fields if `unused`.

In that case I imagined the flag to be parameterized on two aspects fail/warn on unused and whether to track all fields or none of the fields by default.

```scala
// TODO: does this make sense? isn't it too verbose?
.enableUnusedFieldPolicy(FailOnUnused, TrackAll)
.enableUnusedFieldPolicy(WarnOnUnused, TrackNone)

// TODO: to manually track a bunch of fields
.withTrackUnusedFields(_.b, _.c)
```

MateuszKubuszok:
Hmm, the way I see it we have 2 possible policies for now:

 * ignore any unused field
 * fail the compilation on any field which was not explicitly marked as unused

and both would additionally fail the compilation if someone marked a field as unused but it was required.

In some more-or-less distant future, after #115, we might have additional policies:

 * ignore unused fields in fallback, failing on unused fields in the main value
 * fail the compilation for for any field, either in main value or fallback, what wasn't explicitly marked as unused
   * this could be problematic as I cannot think of a good DSL to mark such field as explicitly unused

Additionally, we have to keep in mind that:

 * even though we say `field`, actually right now Chimney makes no distinction between vals defined like:
   ```scala
   case class Foo(a: Int, b: String)
   ```

   and:
 
   ```scala
   case class Foo() {
     val a: Int = 0
     val b: String = "text"
   }
   ```

   and, some people would probably prefer to make the distinction which they want to ignore, but that distinction would have to be implemented somewhere in the future, so for now, we would have to keep the distinction in mind when naming things, e.g. by using name like "unused source values"

 * these flags are global, so whatever name we use, should not collide with whatever users' might use in their codebases, verbose names are less of an issue

 * we would have to add similar thing for target subtypes (when it's an enum/a subtype), so we have to make sure that the name indicates which is which

 * ATM Chimney does not support showing warnings - it's either `info` for logs, or `error` for reasons why macro cannot be expanded, we might add it in the future, as something like `policyReportingStrategy(PolicyCheckFailureIsWarning)`/`policyReportingStrategy(PolicyCheckFailureIsError)` but that would increase the scope of this PR too much.
   
---

Currently I think the API what would work the best would be something like:

```scala
// sealed abstract class UnusedFieldPolicy
// object FailOnIgnoredSourceVal extends UnusedFieldPolicy

src.into[Target]
.enableUnusedFieldPolicy(FailOnIgnoredSourceVal) // 1 value for now
.disableUnusedFieldPolicy // ignores unused fields, the default
.withFieldUnused(_.foo.bar.baz) // expect that field is unused:
                                // - fail compilation if it is needed
                                // - do NOT treat field as ignored when UnusedFieldPolicy is enabled
                                // - be na noop when UnusedFieldPolicy is disabled and field is not used
.transform
```

Then we could test it similarly to other flags:

 * in both `TotalTransformerProductSpec` and `PartialTransformerProductSpec`:
 * `group("flag .enableUnusedFieldPolicy")`
   * `test("should be disabled by default")`
   * `test("should inform user which unused fields failed")` 
   * `test("should ignore explicitly unused fields")`
   * `test("should not allow marking required fields as unused")`
 * `group("flag .disableUnusedFieldPolicy")`
  * `test("should disable globally enabled .enableUnusedFieldPolicy")`
 * tests would have to check that nesting case classes would also work

---

Adding:

 * distinction between warnings and errors
 * distinction between case class fields and other `val`s
 * support for UnusedTargetSubtypePolicy

would be a job for another PR, and not necessarily something to do soon. I am starting to think that even this change would require some refactoring to how `TransformationContext` works with regard to `Path`s in source, so I might need to do some refactoring, but I am short on time these days and I could do it :/

nimatrueway:
> Hmm, the way I see it we have 2 possible policies for now:
>
> ...

Thanks for sharing your insights Mateus, I skimmed through your comment twice and we are aligned. I need to carve up sometime on the weekend to hack it. 🚀

MateuszKubuszok:
Hey, once again thank you for your effort, I am sorry if you invested some more time to this, since I needed this functionality for another WIP PR (#614) and I had to started implementing it there. I am not sure how long it will take to finish that other PR, but it would be hard now to switch to another implementation.

nimatrueway:
> Hey, once again thank you for your effort, I am sorry if you invested some more time to this, since I needed this functionality for another WIP PR (#614) and I had to started implementing it there. I am not sure how long it will take to finish that other PR, but it would be hard now to switch to another implementation.

Sorry man, I switched job around the same time as I was working on this PR. No, I'm working in Ruby/Sorbet shop and slogging to catch up to this new world. _(Having to deal with Ruby, I can't put into words how much I missed Scala)_

Best of luck to you!

Issue #611, opened 2024-10-01
Support non-field paths in withFieldRenamed source path


MateuszKubuszok:
Currently, a source path of `withFieldRenamed(_.source.path, _.target.path)` can be only [a sequence of nested fields](https://chimney.readthedocs.io/en/stable/supported-transformations/#wiring-the-constructors-parameter-to-its-source-field), since they are applied to top-level src path.

> you can only use val/nullary method/Bean getter as a source field name

This could be extended to support:

 * `_.matching[Subtype]`
 * `_.matchingSome`
 * `_.matchingLeft`
 * `_.matchingRight`
 * `_.everyItem`
 * `_.everyMapKey`
 * `_.everyMapValue`

This would probably have to change how we're going into `Path` within `TransformerConfiguration` AND update the `originalSrc`.

 * probably, a correct dropping of a `Path`'s prefixes in `RenameFrom(path)` combined with setting `originalSrc` to `src` on: matching/iterating over collection would work, but it's not tested.

MateuszKubuszok:
Released in [1.5.0](https://github.com/scalalandio/chimney/releases/tag/1.5.0)

PR #612, opened 2024-10-02
Refactor TransformerConfiguration Paths to more explicitly trace what is matched on source-side and what on target-side

MateuszKubuszok:
Prerequisite for #248 and #611

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/Configurations.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/Configurations.scala
@@ -173,11 +173,6 @@ private[compiletime] trait Configurations { this: Derivation =>
         path.segments.headOption.collect { case Segment.Matching(tpe) => tpe -> new Path(path.segments.tail) }
     }
 
-    object AtSourceSubtype {
-      def unapply(path: Path): Option[(??, Path)] =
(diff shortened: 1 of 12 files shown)


chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/Configurations.scala
```diff
@@ -173,11 +173,6 @@ private[compiletime] trait Configurations { this: Derivation =>
         path.segments.headOption.collect { case Segment.Matching(tpe) => tpe -> new Path(path.segments.tail) }
     }
 
-    object AtSourceSubtype {
-      def unapply(path: Path): Option[(??, Path)] =
-        path.segments.headOption.collect { case Segment.SourceMatching(tpe) => tpe -> new Path(path.segments.tail) }
-    }
-
```
MateuszKubuszok, line 180: This special-case hack is no longer needed

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/Configurations.scala
```diff
@@ -585,7 +609,7 @@ private[compiletime] trait Configurations { this: Derivation =>
         extractPath[PathType2].matching[Subtype]
       case ChimneyType.Path.SourceMatching(init, sourceSubtype) =>
```
MateuszKubuszok, line 610: We have to leave this in the DSL runtime (bincompat), so we might as well still parse it.

PR #622, opened 2024-10-07
fallback to names in sealed hierarchies stable order

reimai:
A follow up to https://github.com/scalalandio/chimney/pull/543
Turned out it's quite possible to have NoPosition (with 0, 0 coordinates) in proto generated sources, which ruins current sort.


MateuszKubuszok:
I'd report I there https://github.com/scala/scala3/issues/21672, since it might be a bug in the compiler.

As for the fix, I'd prefer something that relies on

```
implicit val ordering: Ordering[Symbol] = Ordering.by(_.pos).orElseBy(_.name)
```

since it's easier to maintain.

MateuszKubuszok:
I wanted to explain how I'd see the fix but it was easier to just write it https://github.com/scalalandio/chimney/pull/623

MateuszKubuszok:
Released in [1.5.0](https://github.com/scalalandio/chimney/releases/tag/1.5.0)

PR #637, opened 2024-11-20
Add TransformerClass to avoid multiple anonymous classes creation during macro expansion.

pawelsadlo:
Scala 3.3.4 added a new warning for inline methods in case an anonymous class is created.
This introduces concrete class and instantiates it instead of creating anonymous class.

### DIFF
--- a/chimney/src/main/scala-3/io/scalaland/chimney/internal/compiletime/ChimneyExprsPlatform.scala
+++ b/chimney/src/main/scala-3/io/scalaland/chimney/internal/compiletime/ChimneyExprsPlatform.scala
@@ -1,16 +1,25 @@
 package io.scalaland.chimney.internal.compiletime
 
+import io.scalaland.chimney.*
 import io.scalaland.chimney.dsl.TransformerDefinitionCommons
-import io.scalaland.chimney.{integrations, partial, PartialTransformer, Patcher, Transformer}
 
 import scala.collection.compat.Factory
 import scala.quoted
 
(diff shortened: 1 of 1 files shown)


MateuszKubuszok:
Hello, thank you for the contribution!

From my point of view this solution is problematic:

 1. first of all, while it shuts the compiler complaints, it still creates an anonymous class - every lambda on JVM is an anonymous class, so it replaced anonymous `Transformer` with a boxed anonymous `Function1` (`new Transformer$1 {}` -> `new TransformerClass(new Function1$anon$1 {}`, maybe even with some runtime overhead). Other libraries might have used this approach as they were creating anonymous instanced for something that composed 2 other concrete instances but Chimney will _always_ create a new expression, whether we box it with as a lambda in some class or inline as Single Abstract Method
 2. `compiletime` package is no place for runtime constructs as it isn't checked with MiMa, this code creates dependency from users' runtime on macros' internals which are a moving target (which isn't an issue since runtime has no business calling them, so bytecode mismatch never happens)
 3. since it doesn't actually prevent anonymous class creation, only masks it, it should be possible to be implemented as just a new default to `@nowarn` annotation (https://github.com/scalalandio/chimney/blob/7f77d80b1bda5e819499850f5ee2216b6518a497/chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/GatewayCommons.scala#L73), something like `@nowarn("msg=anonymous")` or similar Scala-3 only workaround, however I cannot verify it since... this errors doesn't show either in our CI tests (on 3.3.4) not in [Scastie](https://scastie.scala-lang.org/mxussfEsR4uYQAlQjrEdFw) when I tried to reproduce it

PR #664, opened 2025-01-13
Update Scala.js and set Scala.js version in docs

MateuszKubuszok:
Replaces https://github.com/scalalandio/chimney/pull/663


MateuszKubuszok:
It seems that missing Scala.js version wasn't the issue :/

PR #679, opened 2025-01-26
Add `Tuple transformations` section to the `ducktape` comparison

arainko:
Adding in comparisons between tuple transformations and more up-to-date usage of  `Mode[F]` (i.e. using `.locally` with a local Mode as opposed to a single toplevel one) 

Screenshots of the new section:
Screenshot from 2025-01-26 13-20-27
Screenshot from 2025-01-26 13-20-35

### DIFF
--- a/docs/docs/troubleshooting.md
+++ b/docs/docs/troubleshooting.md
@@ -1600,7 +1600,6 @@ Here are some features it shares with Chimney (Ducktape's code based on GitHub P
         case Card(digits: Long, name: String)
         case Cash
     
-    @main def example: Unit = {
     val transfer = wire.PaymentMethod.Transfer("2764262")
     
     val wirePerson = wire.Person(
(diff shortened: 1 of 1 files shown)


docs/docs/troubleshooting.md
```diff
@@ -1934,8 +2034,6 @@ Since Ducktape is inspired by Chimney, there is a huge overlap in functionality.
     * customizing the [field-](supported-transformations.md#customizing-field-name-matching) and
       [subtype-name](supported-transformations.md#customizing-subtype-name-matching) matching methods
     * [sharing flags overrides between all derivations in the same scope](cookbook.md#reusing-the-flags-for-several-transformationspatchings)
-    * [smart constructors](supported-transformations.md#types-with-manually-provided-constructors), not only custom
```
arainko, line 1937: Dropping this point since ducktape also supports smart cosntructors (https://github.com/arainko/ducktape/blob/456df5656e223ebab9bcc044f65230481591b6ec/ducktape/src/main/scala/io/github/arainko/ducktape/syntax.scala#L107) - I'll have a follow up PR that fills this in the docs as well :smile:

MateuszKubuszok:
Thanks!

PR #680, opened 2025-01-27
Fix typos in DESIGN.md

civilizeddev:
- DLS → DSL
- what is represents → what it represented
- field od subtype → field or subtype
- flags to flag type representation → flags to the flag type representation
- quotes depends → quotes depend
- bo be designed → to be designed
- to and empty config → to an empty config


MateuszKubuszok:
Thanks!

Issue #683, opened 2025-01-29
Regression with handling protobuf Empty in 1.7.0


Krever:
After upgrading from 1.5.0 to 1.7.0 I noticed that code below doesnt compile anymore (its a minimized version)

```proto
syntax = "proto3";

package swissborg.protobuf.ledger.journal;

option java_package = "com.swissborg.protobuf.ledger.journal";

message Foo {
  oneof sealed_value {
    Foo1 foo1 = 1;
  }
}

message Foo1 {}
```

```scala
import com.swissborg.protobuf.ledger.{journal => proto}
import io.scalaland.chimney
import io.scalaland.chimney.protobufs._

sealed trait Foo
object Foo {
  final case class Bar1() extends Foo

  implicit lazy val sagaCodec: chimney.Codec[Foo, proto.Foo] =
    chimney.Codec.define[Foo, proto.Foo]
      .withSealedSubtypeRenamed[Bar1, proto.Foo1]
      .buildCodec

}
```

It can be further minimized into 
```scala 
    sealed trait Proto extends Product with Serializable
    object Proto {
      case class Foo(a: Int) extends Proto
      case object Empty extends Proto
    }

    sealed trait Domain extends Product with Serializable
    object Domain {
      case class Foo1(a: Int) extends Domain
    }

    implicit def handleEmpty[To]: PartialTransformer[Proto.Empty.type, To] =
      (_: Proto.Empty.type, _: Boolean) => partial.Result.fromEmpty[To]

    (Proto.Foo(10): Proto)
      .intoPartial[Domain]
      .withSealedSubtypeRenamed[Proto.Foo, Domain.Foo1]
      .transform
      .asOption ==> Some(Domain.Foo1(10))
    (Proto.Empty: Proto)
      .intoPartial[Domain]
      .withSealedSubtypeRenamed[Proto.Foo, Domain.Foo1]
      .transform
      .asOption ==> None
```

PR #687, opened 2025-02-01
fix: enable UUID field overrides with patch

hughsimpson:
There seems to have been a regression in the 1.7.x release, where patching a UUID field failed compilation. The problem is not limited to UUIDs, but applies to patches where a field is any pojo class with a single constructor and no matching accessors.

The fix I've proposed here is not an elegant one, but it does seem to work; the basic approach is just to forbid TransformProductToProductRule from applying where From <:< To

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformProductToProductRuleModule.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformProductToProductRuleModule.scala
@@ -22,6 +22,10 @@ private[compiletime] trait TransformProductToProductRuleModule { this: Derivatio
     def expand[From, To](implicit ctx: TransformationContext[From, To]): DerivationResult[Rule.ExpansionResult[To]] =
       // From is checked after To, because extraction always succeeds
       (Type[To], Type[From]) match {
+        case (l, r) if r <:< l && ctx.config.areLocalFlagsAndOverridesEmpty =>
+          DerivationResult.attemptNextRuleBecause(
+            s"Type ${Type.prettyPrint[From]} <:< ${Type.prettyPrint[To]}. Falling back to TransformSubtypesRule"
+          )
         case (HasCustomConstructor(constructorOverride), Product.Extraction(fromExtractors)) =>
           mapOverridesAndExtractorsToConstructorArguments[From, To](fromExtractors, constructorOverride)
         case (Product.Constructor(parameters, constructor), Product.Extraction(fromExtractors)) =>
--- a/chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformProductToProductRuleModule.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformProductToProductRuleModule.scala
@@ -22,10 +22,6 @@ private[compiletime] trait TransformProductToProductRuleModule { this: Derivatio
     def expand[From, To](implicit ctx: TransformationContext[From, To]): DerivationResult[Rule.ExpansionResult[To]] =
       // From is checked after To, because extraction always succeeds
       (Type[To], Type[From]) match {
-        case (l, r) if r <:< l && ctx.config.areLocalFlagsAndOverridesEmpty =>
-          DerivationResult.attemptNextRuleBecause(
-            s"Type ${Type.prettyPrint[From]} <:< ${Type.prettyPrint[To]}. Falling back to TransformSubtypesRule"
-          )
         case (HasCustomConstructor(constructorOverride), Product.Extraction(fromExtractors)) =>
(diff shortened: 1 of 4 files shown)


MateuszKubuszok:
Probably the same problem as with #683 and would require similar fix to #684 - since `Patcher`s are reimplemented as specialized `withFallback`s, fallback issues propagate to `Patcher`s:

 - something prevented subtyping rule to apply
 - subtyping and implicits are skipped without trying only by non-empty overrides
 - so in this case overrides should have been empty but weren't
 - patchers wihout overrides -> non-empty overrides had to be some fallbacks in `TransformationContext`

I'It would have to be debugged with `enableMacrosLogging`, but I probably won't have the time for it anytime soon.

hughsimpson:
Cheers. I'll try to look into this myself in the next couple of days

hughsimpson:
I'm comparing with `Instant` which works, and the rule that triggers for UUID but not Instant is `PatchProductWithProduct`. TBH that rule looks a little sketchy

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformProductToProductRuleModule.scala
```diff
@@ -22,6 +22,10 @@ private[compiletime] trait TransformProductToProductRuleModule { this: Derivatio
     def expand[From, To](implicit ctx: TransformationContext[From, To]): DerivationResult[Rule.ExpansionResult[To]] =
       // From is checked after To, because extraction always succeeds
       (Type[To], Type[From]) match {
+        case (l, r) if r <:< l && ctx.config.areLocalFlagsAndOverridesEmpty =>
+          DerivationResult.attemptNextRuleBecause(
+            s"Type ${Type.prettyPrint[From]} <:< ${Type.prettyPrint[To]}. Falling back to TransformSubtypesRule"
```
hughsimpson, line 27: This is pretty hacky... rationale is we only hit this point when called from `PatchProductWithProductRule` (since those are the only times the step's not sequenced to occur at some point after `TransformSubtypesRule`, coming instead after `PatchSubtypeRuleModule`). A root issue seems to be that `ctx.config.filterCurrentOverridesForFallbacks.isEmpty`, when called from `PatchSubtypeRuleModule`, is always false on non-case-class pojos?. I wasn't able to figure out why this was happening or reliably filter out 'incorrect' overrides there.

hughsimpson:
@MateuszKubuszok this is the best working fix that I could manage on the issue... let me know what you think, or if there's a better solution that strikes you.

chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformProductToProductRuleModule.scala
MateuszKubuszok: This condition `r <:< l && ctx.config.areLocalFlagsAndOverridesEmpty` in normal transformers was caught by [Subtype rule](https://github.com/scalalandio/chimney/blob/2eae39303ca6c1615575a8578250714015776d4b/chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/transformer/rules/TransformSubtypesRuleModule.scala#L12).

But it didn't work with `Patcher`s, so I gad to create dedicated [Subtyping rule for Patchers](https://github.com/scalalandio/chimney/blob/master/chimney/src/main/scala/io/scalaland/chimney/internal/compiletime/derivation/patcher/rules/PatchSubtypeRuleModule.scala), without that some existing `Patcher` test cases was falling. Perhaps, it solved one issue but created/missed another.

It should never have reached the `ProductToProduct` rule if the input is expected to be passed unchanged.

hughsimpson: I get the feeling that the 'proper' fix should be something to do with that rule, but another one would be just to apply `TransformSubtypesRule` before `PatchProductWithProductRule`, which also seems to work, and removes this wart...

hughsimpson: I'll take another look at this tomorrow, thanks for the comments

hughsimpson: Ok so I've removed this and just pushed the rule reordering. I guess the reason the check was required in the first step was because it was before the Option & Collection rules and their overrides, but by the time you get to product, they've been resolved. This feels like it's probably a neater fix than writing some more special casing... If I can muster the energy to do so I might try to figure out a way to get the first subtype to match when it's supposed to but that won't be for a few days at least

MateuszKubuszok:
If it pass all tests, then it's good. :) Thank you!

hughsimpson:
Brilliant! Thanks for the prompt review and release! :pray:

PR #713, opened 2025-03-15
Rewrite API so that Scala 3 would not rely on AutoDerived pattern (Scala 3.7 givens resolution change)

MateuszKubuszok:
This PR solves [the issue](https://github.com/scalalandio/chimney/discussions/592) caused by [3.7 givens resolution change](https://www.scala-lang.org/2024/08/19/given-priority-change-3.7.html):

 * it no longer relies on this pattern

   ```scala
   trait TypeClass[A] extends TypeClass.AutoDerived[A]
   object TypeClass {
     trait AutoDerived[A]
     object AutoDerived {
       // summons TypeClass but NOT TypeClass.AutoDerived
       inline given [A]: AutoDerived[A] = ...
     } 
   }
   extension [A](value: A) def foo(using TypeClass.AutoDerived[A]) = ...
   ```
   for recursive (semi)automatic derivation

 * it uses new [Expr.summonIgnoring](https://github.com/scala/scala3/discussions/21909)
 
   ```scala
   trait TypeClass[A]
   object TypeClass {
     inline given derived[A]: AutoDerived[A] =
       ${ macroUsingSimmonIgnoring[A] }
       // Inside it uses:
       // Expr.summonIgnoring[TypeClass[A]](
       //  Symbol.classSymbol("TypeClass").companionModule.methodMember("derived")*
       // )
       // instead of Expr.summon[TypeClass[A]]
   }
   extension [A](value: A) def foo(using TypeClass[A]) = ...
   ```

---

- [x] remove `AutoDerived` on Scala 3
- [x] use `summonIgnoring` on Scala 3
- [x] fix failing tests
  - [x] fix `Type.prettyPrint` on Scala 3 for internal classes (now they have `_` prefix)
  - [x] remove auto/syntax/inline on Scala 3 (they no longer make sense) and make their tests Scala 2-only
  - [x] tests on Scala.js are failing due to https://github.com/scala/scala3/issues/22794 - unblocked
- [x] fix warnings and re-enable `Xfatal-errors` on Scala 3
  - [x] currently warnings are disabled because of https://github.com/scala/scala3/
issues/22812 - unblocked
  - [x] fix warnings on chimney
  - [x] fix warnings on chimney-cats
  - [x] fix warnings on chimney-java-collections
  - [x] fix warnings on chimney-protobufs
- [x] update documentation (differences between Scala 2 and 3 in 2.0.0)
  - [x] update documentation on `master` to explain issues with 3.7 and provide a workaround ([`into.transform` instead of `transformInto`, or using macros](https://scastie.scala-lang.org/MateuszKubuszok/y6ITFRjlQk6li4AWwD24xw/9) seems to work!)


chimney/src/main/scala-3/io/scalaland/chimney/PartialTransformer.scala
```diff
@@ -168,20 +169,11 @@ object PartialTransformer extends PartialTransformerCompanionPlatform {
...
-    def transform(src: From, failFast: Boolean): partial.Result[To]
-  }
-
-  /** @since 0.8.0 */
-  object AutoDerived extends PartialTransformerAutoDerivedCompanionPlatform {
-
-    implicit def liftTotal[From, To](implicit total: Transformer[From, To]): AutoDerived[From, To] =
-      (src: From, failFast: Boolean) => partial.Result.fromCatching(total.transform(src))
-  }
+  type AutoDerived[From, To] = Transformer[From, To]
```
ghostdogpr, line 172: Shouldn't this be `PartialTransformer[From, To]`?

MateuszKubuszok: Well, yes, it should

PR #742, opened 2025-05-23
Prioritize value class unwrapping rules over Option rules

### LINKED ISSUE #741: `PartialTransformer` fails at runtime for nested type transformations involving `Option.None`
**Checklist**

- [x] I read the documentation at https://chimney.readthedocs.io/ and checked that the functionality exists
- [x] I verified that the behavior for my use case doesn't match the documentation
- [x] I checked the https://github.com/scalalandio/chimney/issues and haven't found the issue reported
  - [x] I checked the https://github.com/scala/bug/issues and haven't found the issue reported there - relevant for Scala 2 compiler crashes, since it might be a bug in the compiler
  - [x] I checked the https://github.com/lampepfl/dotty/issues and haven't found the issue reported there - relevant for Scala 3 compiler crashes, since it might be a bug in the compiler 
- [x] I confirmed that the bug is not related to functionality that was deprecated: lifted transformers (`TransformerF`s) or `unsafeOption` flags
- [x] I believe that this is actual bug, rather than [a question how to use existing features](https://github.com/scalalandio/chimney/discussions)

**Describe the bug**

`PartialTransformer` fails unexpectedly at runtime for transformations that are actually total and safe (i.e. there is no Option unwrapping here).

**Reproduction**

I can only reproduce the failure with the nested None mapping. With Some it seems to work fine.

```scala
//> using jvm 21.0.6
//> using scala 2.13.16
//> using dep io.scalaland::chimney::1.8.0

object Main extends App {
    import io.scalaland.chimney.Codec

    case class Item(id: String, data: String)

    case class GItem(id: String, data: String)
    case class GItemOptional(entry: Option[GItem])

    val codec = Codec
        .define[Option[Item], GItemOptional]
        .enableNonAnyValWrappers
        .buildCodec

    println(codec.decode
        .transformFailFast(GItemOptional(Some(GItem("id", "xyz"))))
        .asOption
        .get) // ok; prints Some(Item("id", "xyz"))

    println(codec.decode
        .transformFailFast(GItemOptional(None))
        .asOption
        .get) // fails with java.util.NoSuchElementException: None.get; should print None
}
```

Also, trying to derive an `Iso` in this example should work, but fails.

**Expected behavior**

Be able to build an `Iso` and a working `PartialTransformer`.

**Actual behavior**

Fail at runtime. Even if this is a bug in the total derivation, what is the rationale for failing at runtime without any `Option` unwrapping? (Or, is it being unwrapped unexpectedly...)

**Which Chimney version do you use**

1.8.0.

**Which platform do you use**

- [x] JVM
- [ ] Scala.js
- [ ] Scala Native

OpenJDK Runtime Environment Temurin-21.0.6+7 (build 21.0.6+7-LTS)

Let me know if I'm missing anything here.

bdmendes:
Fixes #741. A simple reordering of rules sufficed.

### DIFF
--- a/chimney/src/main/scala-2/io/scalaland/chimney/internal/compiletime/derivation/transformer/DerivationPlatform.scala
+++ b/chimney/src/main/scala-2/io/scalaland/chimney/internal/compiletime/derivation/transformer/DerivationPlatform.scala
@@ -15,12 +15,12 @@ trait DerivationPlatform
     with rules.TransformSubtypesRuleModule
     with rules.TransformTypeConstraintRuleModule
     with rules.TransformToSingletonRuleModule
-    with rules.TransformOptionToOptionRuleModule
-    with rules.TransformPartialOptionToNonOptionRuleModule
-    with rules.TransformToOptionRuleModule
     with rules.TransformValueClassToValueClassRuleModule
     with rules.TransformValueClassToTypeRuleModule
     with rules.TransformTypeToValueClassRuleModule
+    with rules.TransformOptionToOptionRuleModule
+    with rules.TransformPartialOptionToNonOptionRuleModule
+    with rules.TransformToOptionRuleModule
     with rules.TransformEitherToEitherRuleModule
     with rules.TransformMapToMapRuleModule
     with rules.TransformIterableToIterableRuleModule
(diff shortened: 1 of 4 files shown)


MateuszKubuszok:
Thanks! I'm glad that change of priorities was enough. I'll release it after weekend.

bdmendes:
Awesome! Thanks for the support.

PR #759, opened 2025-07-06
Add test for the fix of #758

### LINKED ISSUE #758: Error when deriving for exported Type.
> 
> 
> 
> Is it possible to minimize the reproduction? Ideally as Scala CLI script?
> 
> 
> Hello I have a similar problem here. I don't know much about scala cli script but i make a minimal reproduction here.
> 
> main.scala
> ```
> import hello.*
> 
> import io.scalaland.chimney.dsl.*
> import io.scalaland.chimney.Transformer
> 
> 
> case class UserName(name: String)
> 
> @main
> def main(): Unit =
>   val username = UserName("BBB")
>   val user = username.into[User].withFieldComputed(_.address, _ => "Address").transform
>   println(user)
> 
> ```
> 
> 
> hello.Hello.scala
> ```
> package hello
> 
> //type AAA = World.AAA
> //val AAA = World.AAA
> 
> export World.User
> ```
> 
> hello.World.scala
> ```
> package hello
> 
> class World {
> 
> }
> 
> object World {
>   case class User(name: String, address: String)
> }
> ```
> 
> After compilation i got these errors:
> 
> ```
> Chimney can't derive transformation from UserName to hello.World.User
> 
> hello.World.User
>   derivation from username: UserName to hello.World.User is not supported in Chimney!
> 
> Consult https://chimney.readthedocs.io for usage examples.
> 
> 
>   val user = username.into[User].withFieldComputed(_.address, _ => "Address").transform
> 
> ```
> 
> The another way to export a case class don't work as well.
> 
> I guess this is because chimney just don't work with export tricks in scala. Is this a known limitation? 
> 
> 
> 
>  

 _Originally posted by @ETCHKILI in [#673](https://github.com/scalalandio/chimney/issues/673#issuecomment-3034775890)_

A duplicate of this, to distinguish it from the original issue in that thread.

ETCHKILI:
Fix #758.
Add test for [scalalandio/chimney-macro-commons/#21](https://github.com/scalalandio/chimney-macro-commons/pull/21)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -27,7 +27,7 @@ val versions = new {
   val platforms = List(VirtualAxis.jvm, VirtualAxis.js, VirtualAxis.native)
 
   // Dependencies
-  val macroCommons = "afb290c1cd18210ca4c1cdd95d82d74ec3c58435-SNAPSHOT"
+  val macroCommons = "2.0.1"
   val cats = "2.13.0"
   val kindProjector = "0.13.3"
   val munit = "1.1.1"


ETCHKILI:
Hi  @MateuszKubuszok, i have added some test to the fix. 
I run all the tests (including the markdown tests and test suites) locally.
Mostly they all passed, but sometimes there are some test timeout errors, which I guess it's because of my computer performance.
If there are any other issues here, I'd be happy to continue solving them.

ETCHKILI:
> Just bump the version and we can merge it 🙂

Done! 🥳

PR #870, opened 2026-04-16
Tests for issue #835. 

ajozwik:
Test for issue #835 
It needs [scalalandio/chimney-macro-commons#85](https://github.com/scalalandio/chimney-macro-commons/pull/85) to be merged and published first.


MateuszKubuszok:
Pulled as a part of #872

PR #871, opened 2026-04-18
add ignore prefix names comparison

YannMoisan:
Adds a new `IgnorePrefix` name matching strategy to `TransformedNamesComparison`, letting users match source names to target names by stripping a shared prefix from the source side.

## Motivation

Protobuf-generated enums (and similar code-generated types) commonly prefix every variant with the enum name to avoid collisions at the package level — e.g. `FooBar`, `FooBaz` for an enum `Foo`. See [Buf's "enum names need prefixes" post](https://buf.build/blog/totw-3-enum-names-need-prefixes) for the rationale.

When transforming such generated types into domain ADTs whose cases are written without the redundant prefix (`Bar`, `Baz`), neither `StrictEquality` nor `CaseInsensitiveEquality` matches. Users currently have to write a custom `TransformedNamesComparison` for every such enum.

## Changes

- New `sealed abstract class IgnorePrefix(prefix: String) extends TransformedNamesComparison`
  in `TransformedNamesComparison.scala`.

## Usage

```
case object FooNamesComparison extends IgnorePrefix("Foo")
```

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/dsl/TransformedNamesComparison.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/dsl/TransformedNamesComparison.scala
@@ -49,6 +49,19 @@ object TransformedNamesComparison {
     def namesMatch(fromName: String, toName: String): Boolean = fromName.equalsIgnoreCase(toName)
   }
 
+  /** Matches Strings ignoring given prefix.
+    *
+    * usage: case object FooNamesComparison extends IgnorePrefixNamesComparison("Foo")
+    *
+    * motivation: https://buf.build/blog/totw-3-enum-names-need-prefixes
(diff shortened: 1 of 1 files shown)


MateuszKubuszok:
Thanks Yann! I've folded this into #912, which adds `IgnorePrefix` with you as the commit author (and me as co-author of the fixups). A few adjustments there: dropped `sealed` so downstream code can actually extend it (the documented `case object Foo extends IgnorePrefix("Foo")` usage wouldn't compile otherwise — a test object in a separate file now guards this), replaced the usage/motivation comment with a one-line Scaladoc matching the sibling comparisons, and added unit tests plus a user-facing entry in `supported-transformations.md`. Closing in favour of #912.

PR #882, opened 2026-05-20
Implementation for mapper for snake case to camel case #831

ajozwik:
Implementation for mapper for snake case to camel case described in #831, and fix documentation issue (spaces missing in cookbook)

PR #883, opened 2026-05-20
Fix missing spaces

ajozwik:
Fix missing spaces, and used to check code coverage without changing of scala code

### DIFF
--- a/docs/docs/cookbook.md
+++ b/docs/docs/cookbook.md
@@ -233,9 +233,9 @@ Similarly `Patcher` has `.withPatchedValueFlag(pathFromPatchedValue)`:
 
 ## Checking for unused source fields/unmatched target subtypes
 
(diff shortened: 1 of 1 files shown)


ajozwik:
This PR corrects spaces in : `Converting from Cats collections` section (and two others).
Trigger build was used to check, if indirect coverage of PR for #831 is correct (adding Implementation of mapper for snake case to camel case).

MateuszKubuszok:
Superseded by #912. This is a strict subset of the `cookbook.md` changes in #882, which #912 already incorporates — but note that de-indenting the `Scala 2`/`Scala 3` labels inside the first `!!! example` breaks its rendering, so #912 keeps them indented while retaining the genuine typo/indentation fixes. Closing in favour of #912.

Issue #316, opened 2023-05-04
Migrate `expandSealedClasses` derivation rule to new architecture


MateuszKubuszok:

MateuszKubuszok:
Merged to `scala-3` in #320.

PR #903, opened 2026-07-03
Migrate macro backend from chimney-macro-commons to Hearth 0.4.0

MateuszKubuszok:
Migrates Chimney's macro backend from `chimney-macro-commons` 2.1.0 to **Hearth** (currently `0.4.0-27-g71200e3-SNAPSHOT` — see the release blocker at the bottom).

## What changed

- **Engine**: the whole derivation engine (foundation, rule modules, patcher, Gateways, DSL macros) now runs on Hearth — `MacroCommons`, **cross-quotes** (single-source `Type`/`Expr` construction for both Scala versions), `MIO` (replacing `DerivationResult`; error ADTs are stackless exceptions in `MErrors`; the macro entry point uses Hearth's blessed `runToExprOrFail`), `ValDefs`/`ValDefBuilder`/`LambdaBuilder`/`MatchCase` (replacing `ExprPromise`/`PrependDefinitionsTo`). The `src/main/scala-2`/`scala-3` split shrank from 36 engine files to macro-binding shims + entry classes + `DslMacroUtils` (Scala 2 whitebox stays platform-specific by design).
- **Cross-quotes ownership contract**: instance building was refactored to Hearth's `ValDefBuilder.ofDefN` recipe shape — a spliced `Expr` is constructed inside the scope that splices it (deferred via defs/lazy/MIO), so no owner-healing is needed.
- **`ChimneyMacroExtension` SPI (new)**: an engine-aware, `ServiceLoader`-registered extension point (`io.scalaland.chimney.integrations.ChimneyMacroExtension`) that lets integration libraries register **pair-specific** transformation handlers with access to the derivation engine (recursive `deriveInner`, an outer-map primitive `specialCasedOuterMap`, total/partial context, config/flags). It sits below the implicit rules and above the built-in structural rules. This is the mechanism for conversions that Hearth's shape providers (`IsCollection`/`IsValueType`) can't express (N inner types, total/partial asymmetry, multiple partners per type). `private[chimney]` pending sign-off of the public API shape.
- **Extension layer**: Hearth std providers and `StandardMacroExtension`s are consulted as a fallback: user implicits > `integrations.*` implicits > `ChimneyMacroExtension` handlers > chimney built-ins > Hearth providers (proven by precedence tests).
- **Scala versions / JDK baseline**: Scala 3 → **3.8.4** (sandwich fixture pinned to 3.7.3, sbt-8728); 2.13 → JDK 11+ (`-release 11`), Scala 3 → JDK 17+ (`-release 17`); CI matrix updated.
- **chimney-java-collections**: implicits deleted; module is an unpublished test-only proof that `java.util.*` (+ `Optional`, boxed primitives) derive import-free on JVM. Same-compilation-unit `EnumSet`/`EnumMap` now work.
- **chimney-cats**: collection integrations moved off implicits; **every 1.x use case is preserved** (see "Cats use-case restoration" below).
- **chimney-protobufs**: ships a `ChimneyMacroExtension` (Duration family + `Empty`, engine-aware) plus `StandardMacroExtension` providers (ByteString, 9 wrapper types, Timestamp — import-free, incl. transitive bytes↔BytesValue). oneofs/`UnrecognizedEnum`/`DefaultValue` stay as implicits (documented per-case rationale).
- **Docs**: quickstart/cookbook/under-the-hood/troubleshooting rewritten; new "Hearth macro extensions" guidance for library authors.

## Cats use-case restoration

An early revision of this branch delivered cats collection support only through Kindlings and **dropped** the use cases Kindlings didn't cover. That was corrected: *"move to Kindlings"* governs **how** support is delivered (classpath-only, no implicit import), not **whether** a 1.x use case is supported. Every previously-supported conversion works again, restored **without implicits**:

- **Total `NonEmpty ↔ NonEmpty`** (NonEmptyChain/Vector/Map/Set), **`NonEmptyMap`/`NonEmptySet`** (via a summoned `cats.kernel.Order`), and **`F[A] → G[B]` given `F ~> G` + `Traverse[F]`** (FunctionK) → the new `CatsChimneyMacroExtension` (four `SpecialCaseHandler`s; higher-kinded instances summoned via `Type.decompose1`/`CtorK1`/`summonImplicit`).
- **`NonEmptySeq` / `NonEmptyLazyList`** → `kindlings-cats-integration` `0.3.0-24` (kindlings#163 added the providers).

cats TYPECLASS instances for chimney types + `asValidated*`/`asResult` syntax stay in chimney-cats. Only drift vs 1.x: the empty-input error message is `"Cannot create <TypeName> from empty collection"` (was `"empty value"`); error PATHS unchanged. `NonEmptySet`/`NonEmptyMap` require a `cats.Order` at expansion time (documented). `Either ↔ Validated` remains unsupported (no engine `IsEither` hook — same as 1.x).

## Verification

Local, both Scala versions (JVM), plus JS/Native and docs snippets on CI:

| Module | Scala 2.13 | Scala 3 |
|---|---|---|
| chimney | 980 | 1118 (+4 ignored) |
| chimney-cats | 312 | 312 |
| chimney-protobufs | 39 | 39 |
| chimney-java-collections (tests-only) | 70 | 70 |
| sandwich (2.13↔3) | 6 | 7 |
| JS + Native | green | green |
| docs snippets | all green | — |

CI on the merged head is green except `codecov/patch` + `codecov/project` (threshold noise: covered lines rose, instrumented lines rose faster). MiMa intentionally disabled branch-wide pre-migration (baselines emptied until 2.0.0 GA). Compile-time smoke (benchmarks module, single-shot wall-clock): ~10–20% slower than the macro-commons baseline — full JMH comparison not run.

## Hearth findings — status

24+ Hearth issues were filed during the migration (#307–#344, plus enhancement asks #328–#330, and kindlings #160/#161/#163). Most were fixed upstream and their chimney workarounds **swept**. Notable:

- **Fixed & swept** on Hearth `0.4.0-27`: `Expr.annotated` annotation API (#334 — `nowarn`/`SuppressWarnings`), `DestructuredExpr.Lambda.Param.declaredTpe` for Java-enum value recovery (#341), Scala-2 `CtorN.unapply` + Scala-3 cross-quotes-plugin rewrite of the 3-select `Type.CtorN.UpperBounded.of`/`Bounded.of` (#307/#344 — `ctorNUpperBoundedCompat` removed, ~57 sites), plus the earlier round (#308–#328 introspection/cross-quotes fixes).
- **Remaining live workarounds** (candidates for follow-up Hearth fixes): **#321** (EnumSet provider factory's wildcard `classOf[EnumSet[?]]` fails Scala-2 re-typecheck) and **#333** (`Type.classOfType[IArray[…]]` throws instead of returning `None`), both in `TotallyBuildIterables`.
- **Documented-by-design** (`#342`): `MIO.timeoutDeadlineNanos` save/restore in `GatewayCommons` around `withMioTimeout` (which throws on re-entry — Hearth documented this contract rather than adding a nesting API).
- **Intentional chimney choices** (not workarounds): `reapplyLeadingTypeArgsCompat` (#312, chimney's wildcard-example mechanism), `dealiasedType` (#315, deliberate alias-transparent normalization), non-implicit lazy `Type` vals (#316, style), cross-unit runtime-helper pattern (#320, test-only).

## ⚠️ Release blocker (before 2.0.0 GA)

This branch depends on **`-SNAPSHOT`** artifacts — `hearth 0.4.0-27-g71200e3-SNAPSHOT` and `kindlings-cats-integration 0.3.0-24-gfc36d68-SNAPSHOT` — resolved via a temporary Central-snapshots resolver (and a `COURSIER_REPOSITORIES` line in the snippets CI job). Before releasing 2.0.0, both must be bumped to proper **released** versions and the snapshots resolver + CI line removed. Merged into `2.0.0-development` as an integration step; the snapshot→release bump lands once Hearth cuts a release carrying the #334/#341/#344 fixes.


MateuszKubuszok:
Merged `master` to fix the JS build: sbt-scalajs 1.22.0 + dependency bumps. JS test suites now fully pass with linking (chimneyJS3 1081/1081, chimneyJS 941/941).

The merge also required porting master's feature work onto the Hearth engine (the old-engine implementations landed in deleted paths): **#810 forAll scoped overrides**, **failFast-aware partial methods** (3 runtime types collapse into `ComputedPartialFailFast`/`ConstructorPartialFailFast` under the unified overrides encoding), **#857** (scala-3 intersection types → bounded quoted type variables), and **#835** — whose fix lived in macro-commons 2.2.0, invisible in chimney's own diff; re-implemented as `inheritedFieldGettersCompat` (new Hearth upstream finding: `Type[A].methods` misses inherited parent-class `val` fields on Scala 3). Master's deprecated `withCoproductInstancePartial` was dropped per 2.0.0 fix-by-deletion.

Post-merge totals: chimney 978/978, chimney3 1116/1116 (+4 ignored), cats 312+301, protobufs 36×2, java-collections 70×2, sandwich 6+7, snippets 373/373.

MateuszKubuszok:
CI is fully green as of 8d296da6 — all 13 build jobs (2.13/3 × jvm/js/native × min+latest JDK), formatting, snippets, readthedocs. The earlier JDK-11 failure was the `-release 17` floor applied to the unpublished `chimney-sandwich-test-cases-3` fixture, which the 2.13 job compiles on temurin:11; the fixture is now exempt (verified on local JDK 11.0.12).

The two codecov statuses are structural threshold noise, not a coverage regression: absolute covered lines **increased** 5,884 → 7,000, but instrumented lines grew more (6,732 → 8,323 — the single-sourced Hearth engine is fully instrumentable where the old scala-2/scala-3 split largely wasn't), so the ratio dipped 87.4% → 84.1%.

MateuszKubuszok:
Review round 2 addressed — all 25 points, in four commits (net **−1,132 lines** on top of the migration):

**fa58f97c — Scala 3.8.4 + shared CatsDataSpec** (points 1–2). `-Werror` on Scala 3; sbt-8728 discovery: Scala 3.8's unified `scala-library` coordinate ends the 2.13↔3.8 sandwich, so the sandwich fixture is pinned to 3.7.3 (2.13's tasty-reader ceiling anyway, documented in cookbook). kindlings un-gated on Scala 3 — `CatsDataSpec` is one shared spec, 12/12 identical on both versions.

**079e2230 — expression-level compat dissolved** (3, 6–8, 10, 11, 20–23, 25). `blockExpr`-style combinators, `ScalaExpr`, hoisted helper defs, ambient `Type` implicits, `SmartConstructorResults`, `FactoryCompat` — all gone; cross-quotes inlined at use sites; `fromUntyped` only at genuine SPI boundaries. chimney-protobufs providers now emit self-contained code (`ProtobufsConversions` deleted — the jar works compile-only for the extension mechanism). Custom nowarn machinery replaced by Hearth's built-in unused suppression (the user-facing `@SuppressWarnings` wrapper remainder is hearth#334). Comment sweep per point 25.

**46f1bfac — datatypes as thin Hearth wrappers** (4, 12–16). `IterableOrArrays` **deleted** — `IsCollection`/`IsMap` is the single path for built-ins and extensions, and total-collection codegen switched from iterator+factory to foreach+builder (arrays use the provider's index loop). ProductTypes 926→660 over `Class`; ValueClasses consults `IsValueType` first for everything; singletons broadened (plain objects, Enumeration values — new capability, tested). Deliberate semantics drift per point 12 is catalogued in the commit; every surviving shim cites a hearth issue.

**b3ccb970 — MIO/MErrors direct + DSL single-sourcing + MacroSuite** (9, 17–19, 24). `DerivationResult`/`DerivationErrors` aliases deleted; `ResultSyntax` = logs + `orElseOpt`. DSL parsing single-sourced on `DestructuredExpr` (`DslMacroUtils` deleted on both platforms; platform-specific DSL LOC 5,323→3,769; the S2 whitebox `javaEnumFixedSubtype` stays platform-side — it reads `TypeTree.original` syntax typed exprs don't preserve). `ChimneySpec` is one line extending `hearth.MacroSuite`.

**Upstream**: 24 hearth issues (#307–#328, #333, #334 — bugs decomposed by root cause + enhancement asks) and kindlings#161 filed with minimal repros and workaround links; every surviving workaround in the code cites its issue, ready to sweep when a fixing SNAPSHOT lands. Two findings were not filed (already fixed on hearth master). kindlings#160 was mis-filed and closed — the TASTy mismatch was ours to fix via the 3.8.4 bump.

Full matrix after each wave: chimney 980/980, chimney3 1118/1118 (+4 ignored), cats 312×2, protobufs 36×2, java-collections 70×2, sandwich 6+7, JS3 1083 (+2), snippets 373/373.

MateuszKubuszok:
Swept the workarounds against **hearth 0.4.0-9-g8153a5e-SNAPSHOT** (621738ff, net −436 lines). ⚠️ `versions.hearth` must return to a released hearth before merging — loud TODO in build.sbt.

**Removed** (fix verified by our tests): shims for #308–#314, #316, #319, #320, #322–#324, #326, most of the EnumSet/EnumMap compat (−240 lines), `traverseValDefsCompat`, and the #333 guard (chimney no longer calls `Type.classOfType` at all).

**Restored after failed removal** — the fix doesn't fully cover our case; specifics posted on each issue:
- **#307**: fixed codegen `.dealias.widen`s extracted args → literal String types decay to `java.lang.String` (breaks DSL overrides on Scala 2)
- **#315**: constructor lookups dealias now, but `isClass`/`isAbstract` still read the alias symbol (`export` aliases regress issue 758)
- **#327**: inherited parent vals are listed now, but with `isAvailable=false` (candidate filter drops them, regresses issue 835) and `isInherited=false`

**Kept**: #317/#318 compats — both still reproduce on the snapshot with the exact 0.4.0 signatures (standalone repros re-run; still-reproduces comments posted), and the #334 nowarn bridges.

**Capability gained**: same-compilation-unit-enum `EnumMap` now derives on both Scala versions — the compileErrors pins became real conversion assertions.

Full matrix green (980/1118/312×2/36×2/70×2/6+7/JS3 1083); snippets green (while on the snapshot, snippet runs need the Central snapshots repo in `COURSIER_REPOSITORIES` — dev-only).

MateuszKubuszok:
Sweep round 2 against `0.4.0-16-gd4adc1c-SNAPSHOT`: **#327 swept** (~60 lines, inherited-field flags now correct upstream), **#317 half-swept** (chimney now delegates to the new opt-in `withMacroEntryCtx` API; `prependFreshValCompat` still needed — plain `ValDefs.createVal` trips owner mismatch), **#315's fix verified** but the dealias call stays as deliberate alias-transparent parsing (re-documented as style, 16 NamedTuple specs depend on it), **#307 restored** — a third failure signature surfaced (generated `unapply` vs existentially-quantified projections from wildcard-arg DSL members; specifics on the issue), **#318 not collectible yet** — whole-derivation pinning trades the ScopeException for owner mismatches; suggested `closeScope` owner-healing or hearth-internal cache pinning on the issue. Full matrix + snippets green, totals unchanged.

MateuszKubuszok:
Refactored to the cross-quotes usage contract ("a spliced expr must be created inside the expr splicing it" — deferred construction via defs/lazy vals/MIO laziness): Scala 2/3 instance builders converged into one shared `ValDefBuilder.ofDefN` recipe, `prependFreshValCompat`/`withMacroEntryCtxCompat`/derive-first machinery deleted (net −191 lines), trait-level `Expr` caches became recipes, `TypeCache` scope-partitioned. Owner-healing suggestions retracted on hearth#317/#318 with conforming repro shapes posted; conforming repro variants live next to the originals. Full matrix + snippets green, totals unchanged. Remaining shims: #307 (third signature pending upstream), #334 bridges, plus the release-blocker TODO on the hearth version.

MateuszKubuszok:
Two of the three 'blessed practices' corrections landed (the third, Gateway → `runToExprOrFail`, is next):

- **Ambiguity-list sort by `Type.plainPrint`** (correction #4, a380d5bf) — `prettyPrint` for display, `plainPrint` as the sort key; the ANSI-escape-stripping regex is gone.
- **Java-enum DSL single-sourced via `DestructuredExpr`** (correction #5, 946db377) — one shared branch, only reached on Scala 2; the Scala-2 whitebox special and the interim platform hook are deleted. Empirical caveat: `DestructuredExpr.Lambda.Param.tpe` is the *widened* param type on Scala 2, so the `Color.RED.type` singleton isn't in the typed tree and `Type.Ctor` matching can't recover it — filed [hearth#341](https://github.com/kubuszok/hearth/issues/341) to expose the declared/syntactic type; the value name is recovered from `UntypedExpr.sourceCode` as an interim, to switch to `Type.Ctor` once #341 lands.

Also: [hearth#307](https://github.com/kubuszok/hearth/issues/307#issuecomment-4883151749) now has a standalone repro + root cause (dual-path unapply miss: wildcard-arg → `ExistentialType` on path 1; structural `==` vs symbol-equal aliased-package prefixes on path 2) and a suggested symbol-comparison fix; chimney's `ctorNUpperBoundedCompat` stays until it lands. Full Java-enum + sandwich suites green both Scala versions.

MateuszKubuszok:
**Gateway migrated to `runToExprOrFail`** (correction #3, 55d7db04) — all 6 macro entry points; no more hand-rolled `unsafe.runSync`. Failure output stays byte-identical (25 error specs unchanged), StackOverflow `-Xss64m` guidance preserved (narrow SOE catch), and **flame graphs now work** through chimney (probed: `-Xmacro-settings:hearth.mioBenchmarkScopes` + `mioBenchmarkFlameGraphDir` → valid `.speedscope.json`; under-the-hood.md caveat removed).

Two notes:
- **Macro-logging journal shape changed** (no test asserts it — the only `enableMacrosLogging` spec is commented out): it now renders in Hearth's tree shape rather than chimney's old `Journal.print` shape, because `LogRendering` is a level-filter-only closed ADT and `runToExprOrFail` doesn't expose `state.logs`. Consistent with adopting Hearth's rendering; shout if you want the old shape back.
- Filed [hearth#342](https://github.com/kubuszok/hearth/issues/342): `runToExprOrFail` can't nest (`withMioTimeout` throws on re-entry), which chimney hits when a macro-dependent transformer summons implicits mid-derivation — worked around by save/restore of `MIO.timeoutDeadlineNanos`.

Next: the ChimneyMacroExtension SPI (`IsChimneySpecialCased(handler)` with recursive-derivation access; protobuf Duration/Empty as first users).

MateuszKubuszok:
**ChimneyMacroExtension SPI landed** (e705aca6) — your headline feature. **API up for your review** (kept `private[chimney]` until you sign off the shape):

```scala
// integrations/ChimneyMacroExtension.scala (public entry, ServiceLoader-registered)
abstract class ChimneyMacroExtension extends hearth.MacroExtension[hearth.MacroCommons & ChimneyEngineExtensionApi]
  // def extend(ctx): Unit  →  ctx.registerSpecialCase(handler)

// ChimneyEngineExtensionApi (the engine surface handed to extensions)
type SpecialCaseContext[From, To] = TransformationContext[From, To]
type DerivedExpr[A]               = TransformationExpr[A]

trait SpecialCaseHandler {                                    // ~ your IsChimneySpecialCased(handler)
  def apply[From, To](implicit From: Type[From], To: Type[To]): Option[SpecialCasedTransformation[From, To]]
}
trait SpecialCasedTransformation[From, To] {
  def specialCase(implicit ctx: SpecialCaseContext[From, To]): MIO[Option[DerivedExpr[To]]]  // None = yield
}
def registerSpecialCase(handler: SpecialCaseHandler): Unit
def deriveInner[InnerFrom: Type, InnerTo: Type](...): MIO[...]   // defer inner values to the engine (N of them)
// + specialCasedTotal / specialCasedPartial / specialCasedExpr / specialCaseYield helpers
```

Your `(From,To) match { case IsChimneySpecialCased(handler) => handler.specialCase(using ctx) ; case _ => yield }` is realized as: the new `TransformSpecialCasedRule` (slotted **after the 4 implicit rules, before the built-in structural rules**) tries each registered handler's `apply[From,To]`; first `Some` runs `.specialCase(ctx)`; `None` yields.

**Proof / first user:** chimney-protobufs migrated onto it — the Duration family (proto `Duration` ↔ `java.time.Duration`/`FiniteDuration`/`scala.Duration`, i.e. 3 partners + total/partial asymmetry that `IsValueType`'s one-inner-type limit couldn't express) and `Empty` are now handlers; their implicits are deleted. Any-`To` oneof/UnrecognizedEnum instances stay implicit (not pair-specific).

Green: protobufs 39×2, extension-test 5×2 (incl. recursive-inner + precedence: user implicit beats handler), full matrix + JS unchanged. Naming/visibility is yours to steer — tell me if you'd rename `SpecialCase*`/promote to public.

MateuszKubuszok:
**Snapshot sweep round 3** (02825c09) — bumped to hearth `0.4.0-19-g881908a-SNAPSHOT` (now on Central snapshots; 7-char hash per sbt-git, matching the other published versions).

- **#334 swept**: `nowarnExpr` is now a shared `Expr.annotated(expr, new nowarn(...))` — the scala-3 `@experimental Symbol.freshName` reflection hack + `Symbol.newVal`/`AnnotatedType` block and both scala-2 val-quasiquote overrides are gone. `suppressWarningsExpr` is shared too, except one thin per-platform builder remains because `java.lang.SuppressWarnings` (a *Java* annotation) can't be `new`-ed in expression position cross-platform (flagged on #334 — Hearth's own `Expr.annotated` doc example with `SuppressWarnings` doesn't compile).
- **#307 could NOT be swept — new blocker**: the Scala-2 `unapply` round-2 fix works (chimney 2.13 compiles clean with plain `UpperBounded.of`), but the **Scala-3 cross-quotes plugin has no rewrite case for the 3-select `UpperBounded.of`/`Bounded.of`** (only 2-select `.of`), so `@compileTimeOnly` fires 59×. Filed [hearth#344](https://github.com/kubuszok/hearth/issues/344); `ctorNUpperBoundedCompat` stays until that lands.
- **Still open, shims kept**: #341 (Param.tpe → `javaEnumFixedSubtype` sourceCode recovery), #342 (`runToExprOrFail` no-nest → timeout save/restore).

Local matrix green (chimney 980, chimney3 1118+4, cats/protobufs/java-collections/sandwich/JS3). RELEASE BLOCKER unchanged: `versions.hearth` must reach a released Hearth before merge.

MateuszKubuszok:
**Restored the chimney-cats use cases dropped in Phase 5b** (927fbea8) — following the correction that "move to Kindlings" governs *how* support ships (classpath-only, no implicit import), not *whether* a 1.x use case is supported.

Restored WITHOUT implicits, via the engine-aware SPI:
- **New `CatsChimneyMacroExtension`** (ServiceLoader-registered) with four `SpecialCaseHandler`s: general `Traverse[F]` `F[A]->F[B]`, `NonEmptyMap`/`NonEmptySet` via a summoned `cats.kernel.Order`, and FunctionK `F[A]->G[B]` via a summoned `F ~> G`. HK instances are erased to `Any` at the `CatsMacroConversions` runtime-helper boundary and cast back inside (Scala 2 keeps the spliced tree's real `Traverse[F]` type, which `Traverse`'s invariance won't fit into the erased param).
- **New SPI primitive `ChimneyEngineExtensionApi.specialCasedOuterMap`** — the outer-transformer analog (LambdaBuilder + recursive inner derivation + total/partial fold) the single-value `deriveInner` couldn't express.
- `NonEmptySeq`/`NonEmptyLazyList` now derive via `kindlings-cats-integration` 0.3.0-24 ([kindlings#163](https://github.com/kubuszok/kindlings/issues/163)).

Un-pins the three `CatsDataSpec` `BEHAVIOR-GAP(2.0.0)` tests into passing assertions. Local (central-only): chimney 980, chimney3 1118(+4), chimneyCats 312, chimneyCats3 312 — all green. hearth stays 0.4.0-19.

PR #913, opened 2026-07-11
Add map/contramap/imap combinators to Transformer, PartialTransformer, Codec and Iso

MateuszKubuszok:
Re-lands #591 (by @danicheg) on top of the current 2.0.0 backend and completes it per the review feedback there. @danicheg is the commit author; I co-authored the completion.

Closes #591. Resolves the feature request behind it.

## What's added

Combinators are plain `final def`s directly on the type classes (the dedicated `*Ops` trait from the original review is no longer needed now that the trait hierarchy has been flattened; `@FunctionalInterface`/SAM syntax stays intact):

| Type class | Combinators |
|---|---|
| `Transformer[From, To]` | `map(To => To2)`, `contramap(From2 => From)` |
| `PartialTransformer[From, To]` | `map`, `mapPartial(To => Result[To2])`, `contramap`, `contramapPartial(From2 => Result[From])` |
| `Iso[First, Second]` (invariant) | `imapFirst(f)(g)`, `imapSecond(f)(g)` |
| `Codec[Domain, Dto]` (invariant) | `imapDomain(f)(g)`, `imapDto(f)(g)` |

### Design notes (addressing the #591 review)
- **Full symmetry / parity:** both `Transformer` and `PartialTransformer` get `map` *and* `contramap`; `PartialTransformer` additionally gets the partial variants `mapPartial` / `contramapPartial` (input- and output-side).
- **All applicable type classes:** `Iso` and `Codec` are bidirectional, so they get *invariant* `imap*` combinators taking a there-and-back pair `(f)(g)` (cats-style ordering: `f` forward, `g` backward). They're implemented in terms of the `Transformer`/`PartialTransformer` combinators.
- **`failFast` is threaded through** the `PartialTransformer` combinators (test `combinators preserve fail-fast mode` proves it).
- `Patcher` is intentionally excluded — map/contramap doesn't map onto patching semantics.

## Tests
- `TotalTransformerSpec`, `PartialTransformerSpec` (extended), `CodecSpec`, `IsoSpec` — one case per combinator, incl. error-propagation and fail-fast.
- Green on Scala 2.13 and Scala 3.

## Docs
- New cookbook section **"Adapting existing type class instances"** with four runnable examples, all verified through the `scala-cli-md-spec` snippet harness.

## Verification
- `chimney`/`chimney3` `testOnly` for the four specs: 15 passing each.
- `scalafmtCheckAll` clean.
- MiMa: `2.0.0-development` has empty `mimaPreviousArtifacts` (major-version reset), so nothing to check; all additions are source- and binary-additive regardless.

Note on scope: this was rebuilt fresh rather than `git rebase`d — #591's branch is 471 commits behind and predates the Hearth backend migration (#903), so a literal rebase was not viable.

### DIFF
--- a/chimney/src/main/scala/io/scalaland/chimney/Codec.scala
+++ b/chimney/src/main/scala/io/scalaland/chimney/Codec.scala
@@ -21,7 +21,42 @@ import io.scalaland.chimney.internal.runtime.{TransformerFlags, TransformerOverr
   *
   * @since 1.2.0
   */
-final case class Codec[Domain, Dto](encode: Transformer[Domain, Dto], decode: PartialTransformer[Dto, Domain])
+final case class Codec[Domain, Dto](encode: Transformer[Domain, Dto], decode: PartialTransformer[Dto, Domain]) {
+
+  /** Creates a new [[io.scalaland.chimney.Codec]] with the `Domain` side replaced by `NewDomain`, given a bijection
+    * between the two.
+    *
+    * @tparam NewDomain
+    *   the new domain type
+    * @param f
+    *   conversion from the old `Domain` to `NewDomain`
+    * @param g
+    *   conversion from `NewDomain` back to the old `Domain`
+    * @return
+    *   new [[io.scalaland.chimney.Codec]] between `NewDomain` and `Dto`
+    *
+    * @since 2.0.0
+    */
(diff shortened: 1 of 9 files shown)


danicheg:
Wow, nice to see this is happening. Just curious, what made you change your mind about it? :)

